PR #7, opened 2024-11-20
Add from_json method to classes

dstansby:
Adds a abstract base class for objects that can go to/from JSON. I only added the `to_json` so far.

Also adds a basic test against the example metadata given in the v0.4 spec.

### DIFF
--- a/src/ome_zarr_models/v04/image.py
+++ b/src/ome_zarr_models/v04/image.py
@@ -1,15 +1,32 @@
 from collections.abc import Mapping, Sequence
-from dataclasses import dataclass
-from typing import Any, Literal
-
-from ome_zarr_models.zarr_models.v2 import Group
+from dataclasses import dataclass, fields
+from typing import Any, Literal, Self, get_args
 
 AxisType = Literal["time", "space", "channel"]
 
 
+class ToFromJSON:
+    @classmethod
+    def from_json(cls, json: dict) -> Self:
+        attributes = {}
+        for field in fields(cls):
+            if isinstance(field.type, ToFromJSON):
+                attributes[field.name] = field.from_json(json[field.name])
+            # TODO: fix this crime against programming:
+            elif str(field.type).startswith("collections.abc.Sequence"):
+                attributes[field.name] = []
+                item_type = get_args(field.type)[0]
+                for v in json[field.name]:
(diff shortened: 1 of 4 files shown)


src/ome_zarr_models/v04/image.py
```diff
@@ -1,15 +1,22 @@
...
 from dataclasses import dataclass
-from typing import Any, Literal
-
-from ome_zarr_models.zarr_models.v2 import Group
+from typing import Any, Literal, Self
 
 AxisType = Literal["time", "space", "channel"]
 
 
+class ToFromJSON(ABC):
```
d-v-b, line 9: name suggestion: `JSONable`

src/ome_zarr_models/v04/image.py
```diff
@@ -67,19 +85,53 @@ class TranslationTransform:
...
+    https://ngff.openmicroscopy.org/0.4/index.html#multiscale-md
+    """
+
     path: str
     coordinateTransformations: (
         tuple[ScaleTransform] | tuple[ScaleTransform, TranslationTransform | str]
     )
 
+    @classmethod
+    def _from_json(cls, json) -> Self:
```
d-v-b, line 116: This variable name might cause name collisions if we do `import json`

src/ome_zarr_models/v04/image.py
```diff
@@ -1,15 +1,22 @@
...
 
 AxisType = Literal["time", "space", "channel"]
 
 
+class ToFromJSON(ABC):
+    """A class that can serialise to and from JSON."""
+
+    @classmethod
+    @abstractmethod
+    def _from_json(cls, json: dict) -> Self: ...
```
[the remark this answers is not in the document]
dstansby: Depends what we want to external API to be? I would say lets make input/output like this private to start, then when we write the tutorial we can decide what we should make public?

src/ome_zarr_models/v04/image.py
dstansby: Ooh I like

src/ome_zarr_models/v04/image.py
dstansby: ooh yes, I will change

PR #15, opened 2024-11-21
Feat/pydantic

d-v-b:
adds pydantic, using models from https://github.com/JaneliaSciComp/pydantic-ome-ngff and https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/fractal_tasks_core/ngff/specs.py


src/ome_zarr_models/v04/v04/well.py
```diff
@@ -0,0 +1,79 @@
...
+        for image in self.well.images:
+            if image.acquisition is None:
+                raise ValueError(
+                    "Cannot get acquisition paths for Zarr files without "
+                    "'acquisition' metadata at the well level"
+                )
+            if image.acquisition not in acquisition_dict:
+                acquisition_dict[image.acquisition] = []
+            acquisition_dict[image.acquisition].append(image.path)
+        return acquisition_dict
```
d-v-b, line 79:
```suggestion
```

PR #25, opened 2024-11-22
Feat/multiscale tests

d-v-b:
I bring in a lot of the tests and validation logic from pydantic-zarr.
A few notable changes I add here:
- I added `.build` methods to `VectorScale` and `VectorTranslation` that allow you to create the respective transform without explicitly passing in the `type` as a (redundant) keyword argument
- I added a private `_build_transforms` function that creates `tuple[VectorScale] | tuple[VectorScale, VectorTranslation]`, this is a convenience function for making `coordinateTransformations` metadata.
- I add validation functions for transforms and axes, and the corresponding pydantic model type annotations


d-v-b:
I added validation + tests to ensure that the number of `multiscales.axes` matches the dimensionality of the transforms defined in `multiscales.coordinateTransformations` (if present), and separate validation / tests to ensure that the axes match the number of transforms in the per-dataset coordinate transformations

d-v-b:
i think this is ready for review!

d-v-b:
this _should_ resolve https://github.com/BioImageTools/ome-zarr-models-py/issues/33

src/ome_zarr_models/v04/axes.py
```diff
@@ -1,6 +1,17 @@
+from enum import Enum
 from ome_zarr_models.base import Base
 
 
+class AxisType(str, Enum):
```
d-v-b, line 5: do we need an enum or should we use `typing.getargs` and a literal type? I will try the second option

src/ome_zarr_models/v04/coordinate_transformations.py
```diff
@@ -26,7 +27,18 @@ class VectorScale(Base):
     """
 
     type: Literal["scale"]
-    scale: list[float] = Field(..., min_length=2)
+    scale: list[float]
+
+    @classmethod
+    def build(cls, data: Iterable[float]) -> Self:
```
dstansby, line 33: Could we instead give `type` a defualt (`type: Literal["scale"] = "scale"`), and then if you want to construct one of these you can just do `VectorScale(scale=[1, 2, 3])`? I think that's slightly nicer API and avoids us having to write a `build` method.

src/ome_zarr_models/v04/coordinate_transformations.py
```diff
@@ -26,7 +27,18 @@ class VectorScale(Base):
...
+        """
+        Create a VectorScale from an iterable of floats.
+        """
+        return cls(type="scale", scale=tuple(data))
+
+    @property
+    def ndim(self) -> int:
```
dstansby, line 40: Nice, I like this! Where we think it's sensible sprinling useful properties like this will make the library way more useful.

src/ome_zarr_models/v04/coordinate_transformations.py
d-v-b: the argument against giving the type a default value was that this would allow a JSON object without a `type` field to pass validation, even though that JSON object is technically an invalid axis

src/ome_zarr_models/v04/coordinate_transformations.py
```diff
@@ -69,3 +92,32 @@ class PathTranslation(Base):
 
     type: Literal["translation"]
     translation: str
+
+
+def ndim(transform: VectorScale | VectorTranslation) -> int:
```
dstansby, line 97: In the spirit of don't repeat yourself, I'm not mad about having `ndim` defined as both a method on the classes and a function. Was there a particular reason for doing this?

src/ome_zarr_models/v04/coordinate_transformations.py
```diff
@@ -69,3 +92,32 @@ class PathTranslation(Base):
...
+    if hasattr(transform, "scale"):
+        return transform.ndim
+    elif hasattr(transform, "translation"):
+        return transform.ndim
+    else:
+        msg = f"Cannot infer the dimensionality of {type(transform)}"
+        raise TypeError(msg)
+
+
+def _build_transforms(
```
dstansby, line 110: A lazy cmd+f on GitHub doesn't give me any other hits for this function - is it used elsewhere? (this might just be because GitHub does crappy lazy loading of pages these days 😢 )

pyproject.toml
```diff
@@ -3,8 +3,8 @@ name = "ome-zarr-models"
 dynamic = ["version"]
 description = "Data models for OME-Zarr"
 readme = "README.md"
-requires-python = ">=3.11"
-dependencies = []
+requires-python = ">=3.10"
```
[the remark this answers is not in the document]
d-v-b: ah right, this change needs to be reverted

d-v-b, line 6:
```suggestion
requires-python = ">=3.11"
```

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
 from ome_zarr_models.v04.omero import Omero
+from pydantic_zarr.v2 import ArraySpec, GroupSpec
+import zarr
+
+VALID_NDIM = (2, 3, 4, 5)
+NUM_TX_MAX = 2
+
+
+def ensure_transform_dimensionality(
+    transforms: tuple[VectorScale | PathScale]
```
dstansby, line 25: Would be good to extract the type here as a type alias (or type var???) to avoid repetition.

```
TransformType = tuple[VectorScale | PathScale] | tuple[VectorScale | PathScale, VectorTranslation | PathTranslation]
```

src/ome_zarr_models/v04/coordinate_transformations.py
d-v-b: the main reason to have a free-floating function is to support the niche case of using a functional idiom like `map` or `filter` on transformation objects, e.g. `ndims = map(ndim, transforms)`, but maybe this is too niche :)

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
 )
 from ome_zarr_models.v04.omero import Omero
+from pydantic_zarr.v2 import ArraySpec, GroupSpec
+import zarr
+
+VALID_NDIM = (2, 3, 4, 5)
+NUM_TX_MAX = 2
+
+
+def ensure_transform_dimensionality(
```
dstansby, line 24: Thoughts on public vs. private API of these validation functions? My inclintation is to go private to be on the safe side of being able to modify/fix them easily in the future, and then the public API is just using the classes.

dstansby, line 24: Having read this function, to simplify it I would actually decouple the validation from the spec a bit, and write the validation function to check that an arbitrary list of transforms (that therefore have a `.ndim` attribute) have the same number of dimensions.

So:
1. Remove all transforms from the list that are Path transforms
2. If `>= 2` transforms left, check they have same dimensions

I think this will make typing/logic a bit nicer, and also make this a bit more re-usable for whatever comes in coordinate transform changes.

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
+def ensure_scale_translation(
+    transforms: tuple[VectorScale | PathScale]
+    | tuple[VectorScale | PathScale, VectorTranslation | PathTranslation],
+) -> (
+    tuple[VectorScale | PathScale]
+    | tuple[VectorScale | PathScale, VectorTranslation | PathTranslation]
+):
+    """
+    Ensures that the first element is a scale transformation, the second element,
+    if present, is a translation transform, and that there are only 1 or 2 transforms.
```
dstansby, line 61:
```suggestion
    Ensures that
    -  there are only 1 or 2 transforms.
    - the first element is a scale transformation
    - the second element, if present, is a translation transform
```

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
+    """
+
+    if len(transforms) == 0 or len(transforms) > 2:
```
dstansby, line 64:
```suggestion
    if len(transforms) not in [1, 2]:
```

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
+
+    ndims = tuple(tx.ndim for tx in transforms)
+    ndims_set = set(ndims)
+    if len(ndims_set) > 1:
+        msg = (
+            "The transforms have inconsistent dimensionality. "
+            f"Got transforms with dimensionality = {ndims}."
+        )
+        raise ValueError(msg)
+    return transforms
```
dstansby, line 49: This doesn't actually change/mutate the transforms, so maybe get rid of the return?

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
+    if len(transforms) == NUM_TX_MAX:
+        maybe_trans = transforms[1]
+        if (maybe_trans.type) != "translation":
+            msg = (
+                "The second element of `coordinateTransformations` must be a "
+                f"translation transform. Got {maybe_trans} instead."
+            )
+            raise ValueError(msg)
+
+    return transforms
```
dstansby, line 84: Again, maybe get rid of the return here?

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -1,17 +1,144 @@
...
+    - there are only 2 or 3 axes with type `space`
+    - the axes with type `space` are last in the list of axes
+    - there is only 1 axis with type `time`
+    - there is only 1 axis with type `channel`
+    - there is only 1 axis with a type that is not `space`, `time`, or `channel`
+    """
+    axis_types = [ax.type for ax in axes]
+    type_census = Counter(axis_types)
+    num_spaces = type_census["space"]
+    if num_spaces < 2 or num_spaces > 3:
```
dstansby, line 121:
```suggestion
    if num_spaces not in [2, 3]:
```

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -60,10 +256,69 @@ class MultiscaleGroupAttrs(Base):
...
+    multiscales: tuple[Multiscale, ...] = Field(
         ...,
         description="The multiscale datasets for this image",
         min_length=1,
     )
     omero: Omero | None = None
-    _check_unique = field_validator("multiscales")(_unique_items_validator)
+
+
+class MultiscaleGroup(GroupSpec[MultiscaleGroupAttrs, ArraySpec | GroupSpec]):
```
dstansby, line 267:
```suggestion
class MultiscaleZarrGroup(GroupSpec[MultiscaleGroupAttrs, ArraySpec | GroupSpec]):
```
? Or is my suggestion too verbose?

dstansby: Never mind it's not a zarr group is it so this would just be confusing

src/ome_zarr_models/v04/multiscales.py
d-v-b: definitely, we can define + export this union in the transformations module

src/ome_zarr_models/v04/multiscales.py
d-v-b: private sounds good!

d-v-b: good idea! I will implement this

src/ome_zarr_models/v04/multiscales.py
dstansby, line 269: How is this meant to be used? I am a bit confused/not understanding what the purpose of this class is.

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -50,7 +234,19 @@ class Multiscale(Base):
     metadata: Any = None
     name: Any | None = None
     type: Any = None
-    _check_unique = field_validator("axes")(_unique_items_validator)
+
+    @property
+    def ndim(self) -> int:
```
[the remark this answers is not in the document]
d-v-b: i'm not opposed to it, we would basically be designating "dimensional" data structures

src/ome_zarr_models/v04/coordinate_transformations.py
d-v-b: it's now being used in a validator

src/ome_zarr_models/v04/multiscales.py
d-v-b: don't we need the `return` for it to work as a pydantic validator?

src/ome_zarr_models/v04/multiscales.py
d-v-b: this class will model an actual zarr group, i.e. the attributes (which are JSON and ought to contain multiscales etc) and the arrays (which ought to be the ones named by the `path` attribute in `datasets`, and which ought to have shape that's consistent with the number of axes, etc)

d-v-b: this class is also the thing we deserialize a zarr group into

PR #42, opened 2024-11-22
Feat/zarr io

d-v-b:
add more functionality to `MultiscaleGroup`, and a lot of tests!


d-v-b:
I peeled `from_arrays` and `from_array_props` off the `MultiscaleGroup` class and put it in the tests as over-engineered test utilities. If we need these in the library, we an always bring them back.

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -277,7 +285,63 @@ class MultiscaleGroupAttrs(Base):
...
+    for multiscale in attrs.multiscales:
+        for dataset in multiscale.datasets:
+            dpath = "/" + dataset.path.lstrip("/")
+            if dpath in flattened:
+                if not isinstance(flattened[dpath], ArraySpec):
+                    msg = (
+                        f"The node at {dpath} should be an array, "
+                        f"found {type(flattened[dpath])} instead"
+                    )
+                    raise ValueError(msg)
```
dstansby, line 304: Do we want to eagerly error on the first error we find, or collect up a bunch of errors and then raise a message at the end that includes a message with all the errors found? I'd be inclined to collect them up, so the user can fix them all in one go, instead of having to fix > run code > fix > run code... to find and fix them all.

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -277,7 +285,63 @@ class MultiscaleGroupAttrs(Base):
...
+    Check that all the arrays referenced by the `multiscales` metadata have dimensionality 
+    consistent with the number of axes defined in the metadata.
+    """
+    multimeta = data.attributes.multiscales
+    flat_self = data.to_flat()
+
+    for multiscale in multimeta:
+        multiscale_ndim = len(multiscale.axes)
+        for dataset in multiscale.datasets:
+            arr: ArraySpec = flat_self["/" + dataset.path.lstrip("/")]
```
dstansby, line 326: Perhaps if there's a performance cost to looking up the array metadata (which there will be for remote data), then we should think about merging `_check_array_ndim` and `_check_datasets_exist` to reduce the number of accesses?

But for now, I think it's clearer and easier to keep these as seaparate functions.

src/ome_zarr_models/zarr_utils/__init__.py
dstansby, line 1: Not a huge fan of `zarr_utils/__init__.py`, would suggest moving this to either `ome_zarr_models/zarr_utils.py`, or moving `ome_zarr_models/zarr_models` > `ome_zarr_models/zarr` and making this `ome_zarr_models/zarr/utils.py`

tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from zarr import MemoryStore
+
+
+@pytest.fixture
+def store(request: pytest.FixtureRequest) -> MemoryStore:
```
dstansby, line 6:
```suggestion
def store(request: pytest.FixtureRequest) -> Store:
```
If you intend to return other stores in future?

dstansby (commented):
Left some comments - didn't look at the tests, I trust you with them if you're happy with that (shout if you want me to properly review them though)

src/ome_zarr_models/v04/multiscales.py
d-v-b: all of the IO is done before these functions are called. the input to this function is a pydantic model with no handle to the zarr group. basically we do IO in one place (`from_zarr`), and everything after that is working with in-memory data

src/ome_zarr_models/zarr_utils/__init__.py
d-v-b: I think we can't have `ome_zarr_models/zarr` because attempting to invoke `import zarr` in that module will break python's import resolution?

dstansby: `import ome_zarr_models.zarr` should work? But yeah definitely see not wanting a `zarr` namespace anywhere, even if relative 👍

src/ome_zarr_models/v04/multiscales.py
d-v-b: so it's not easy to gather all of the exceptions here, for reasons that I can expound on if needed. I think people might have to live with incomplete "your data is broken" messages unless pydantic changes things about the execution order of validation functions.

d-v-b:
i think this is ready to go, any further thoughts / concerns @dstansby ?

src/ome_zarr_models/v04/multiscales.py
```diff
@@ -16,13 +17,12 @@
     _build_transforms,
     _ndim,
 )
+from ome_zarr_models.v04.omero import Omero  # noqa: TCH001
```
dstansby, line 20: This noqa looks suspicious to me 🙈  - can you either explain it here, remove it, or just open an issue to track checking these later (happy to stick noqa in for development velocity now, as long as we remember to check later!)

dstansby (approved):
One comment, feel free to resolve as you see fit then merge 🚢

src/ome_zarr_models/v04/multiscales.py
d-v-b: ruff doesn't know that pydantic needs the type annotations at runtime, so without importing `Omero` at runtime, models with fields annotated with `Omero` are broken

d-v-b: (and without the noqa, ruff will stick `Omero` in the `if TYPE_CHECKING` block)

PR #43, opened 2024-11-22
Improve Axes code

dstansby:
This fixes partially https://github.com/BioImageTools/ome-zarr-models-py/issues/30 by:

- Creating a list of `Axis`
- Allowing any type for `unit` (the type is not narrowed in the spec, at least in section 3.1)

### DIFF
--- a/src/ome_zarr_models/v04/axes.py
+++ b/src/ome_zarr_models/v04/axes.py
@@ -1,8 +1,8 @@
-from typing import Literal
+from typing import Any, Literal
 
 from ome_zarr_models.base import Base
 
-__all__ = ["AxisType", "Axis"]
+__all__ = ["AxisType", "Axis", "Axes"]
 
 
 AxisType = Literal["space", "time", "channel"]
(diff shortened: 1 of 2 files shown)


src/ome_zarr_models/v04/axes.py
```diff
@@ -17,4 +17,7 @@ class Axis(Base):
 
     name: str
     type: str | None = None
-    unit: str | None = None
+    unit: str | Any | None = None
+
+
+Axes = list[Axis]
```
d-v-b, line 23: Do we want to restrict this type to `list[Axis]`? I think that will result in functions like `_ensure_axis_length` failing for tuples

dstansby: Ooh good point, changed to `Sequence`

d-v-b: we could even think about making it an iterable at some point, but `Sequence` is fine for now

d-v-b (approved):

PR #48, opened 2024-11-22
Finish labels metadata

dstansby:
Fixes https://github.com/BioImageTools/ome-zarr-models-py/issues/35. The spec is very vague, and there are no concrete examples of this metdata in the spec, but I think my interpretation is sensible and the best one.


src/ome_zarr_models/v04/image.py
```diff
@@ -18,6 +19,9 @@ class Image:
 
     def __init__(self, group: zarr.Group):
         self._attrs = MultiscaleGroupAttrs(**group.attrs.asdict())
+        if "labels" in group:
+            if not isinstance(group["labels"], zarr.Group):
+                raise ValidationError("The special group 'labels' is not a zarr group")
```
d-v-b, line 24: this error message is a bit cryptic -- maybe better to say that a group was expected, but an array was found instead?

d-v-b (approved):
lgtm, if we want to support easy bidirectional serialization of labels groups, then we might need to define `LabelsGroup(pydantic_zarr.v2.Group)` somewhere, and then make that type one of the possible elements of `MultiscaleGroup.members`

PR #49, opened 2024-11-22
Finish omero model

dstansby:
Fixes https://github.com/BioImageTools/ome-zarr-models-py/issues/34.

Note that in order to parse the example given in the spec and not lose the extra bits of metadata, I've enabled allowing extra fields in this PR (which I think we decided we should add at some point anyway?)

### DIFF
--- a/tests/v04/test_omero.py
+++ b/tests/v04/test_omero.py
@@ -1,0 +1,28 @@
+import json
+from pathlib import Path
+
+from ome_zarr_models.v04.omero import Channel, Omero, Window
+
+
+def test_load_example_json():
+    with open(Path(__file__).parent / "data" / "omero_example.json") as f:
+        data = json.load(f)
+
+    print(Omero(**data))
+    assert Omero(**data) == Omero(
+        channels=[
+            Channel(
+                color="0000FF",
+                window=Window(max=65535.0, min=0.0, start=0.0, end=1500.0),
+                active=True,
+                coefficient=1,
+                family="linear",
+                inverted=False,
+                label="LaminB1",
+            )
+        ],
+        id=1,
+        name="example.tif",
+        version="0.4",
+        rdefs={"defaultT": 0, "defaultZ": 118, "model": "color"},
+    )
--- a/tests/v04/test_omero.py
+++ b/tests/v04/test_omero.py
@@ -8,7 +8,6 @@ def test_load_example_json():
     with open(Path(__file__).parent / "data" / "omero_example.json") as f:
         data = json.load(f)
 
-    print(Omero(**data))
     assert Omero(**data) == Omero(
         channels=[
             Channel(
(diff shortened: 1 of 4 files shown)


tests/v04/test_omero.py
```diff
@@ -0,0 +1,28 @@
...
+from pathlib import Path
+
+from ome_zarr_models.v04.omero import Channel, Omero, Window
+
+
+def test_load_example_json():
+    with open(Path(__file__).parent / "data" / "omero_example.json") as f:
+        data = json.load(f)
+
+    print(Omero(**data))
```
d-v-b, line 11: do we need to print here?

tests/v04/test_omero.py
```diff
@@ -0,0 +1,28 @@
+import json
+from pathlib import Path
+
+from ome_zarr_models.v04.omero import Channel, Omero, Window
+
+
+def test_load_example_json():
+    with open(Path(__file__).parent / "data" / "omero_example.json") as f:
```
d-v-b, line 8: we should put these JSON documents in our tests module

tests/v04/test_omero.py
dstansby: lol woops

tests/v04/test_omero.py
dstansby: Isn't this the test module? What filepath within the repo are you rpoosing instead?

tests/v04/test_omero.py
dstansby: This is why we do review 😆

tests/v04/test_omero.py
d-v-b: quite right, `Path(__file__).parent / "data" / "omero_example.json"` will resolve to a path inside the tests module, but we could also store the JSON inline as python strings, and then they can be imported without requiring tests to do IO

d-v-b: (another option would be keep the JSON files as JSON files, but make our tests access them via a pytest fixture that does the IO once per test session)

dstansby: The advantage of storing in a .json file instead of an inline string is getting nice linting/syntax highlighting for it. At the moment this is only read in once per session - perhaps we can leave as is for now, and if/when we add more similar tests think about how to de-duplicate the machinery?

d-v-b: +1 to storing JSON as JSON :) and this is not a blocker, we can handle this when we apply a good first-pass normalization of all our tests

tests/v04/test_omero.py
```diff
@@ -0,0 +1,27 @@
...
+                window=Window(max=65535.0, min=0.0, start=0.0, end=1500.0),
+                active=True,
+                coefficient=1,
+                family="linear",
+                inverted=False,
+                label="LaminB1",
+            )
+        ],
+        id=1,
+        name="example.tif",
```
d-v-b, line 24: haha

d-v-b:
re: the extra fields, that's a good and essential addition. I know of at least one OME-Zarr implementation that uses extra metadata fields (`ngff-zarr`)

d-v-b (approved):
lgtm!

PR #50, opened 2024-11-23
Implement bioformats2raw structure

dstansby:
Implements the parts of https://github.com/BioImageTools/ome-zarr-models-py/issues/31 that do not require parsing XML files (🙈) or zarr groups.

I think the intention of the spec is to have this as another "type" of dataset, alongside HCS and Image, even though it is defined in section 3 (not 2) of the spec. This agrees with the [JSON scheme ipmlementaiton](https://github.com/ome/ngff/blob/main/0.4/schemas/bf2raw.schema) being in it's own file for bioformats2raw.

So I've put it in the `ome_zarr_models.v04` namespace, and I think (for now) we are expecting users to know in advance whether they have a HCS, Image, or BioFormats2Raw dataset? I guess we *could* provide a convenience function to try work out which of these an arbitrary dataset is, but that should go in a separate PR

### DIFF
--- a/src/ome_zarr_models/v04/bioformats2raw.py
+++ b/src/ome_zarr_models/v04/bioformats2raw.py
@@ -1,0 +1,16 @@
+from typing import Any, Literal
+
+from pydantic import Field
+
+from ome_zarr_models.base import Base
+from ome_zarr_models.v04.plate import Plate
+
+
+class BioFormats2Raw(Base):
+    """
+    A bioformats2raw zarr group.
+    """
+
+    bioformats2raw_layout = Field(Literal["3"], alias="bioformats2raw.layout")
+    plate: Plate | None = None
+    series: Any | None = None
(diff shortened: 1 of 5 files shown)


d-v-b:
> I guess we _could_ provide a convenience function to try work out which of these an arbitrary dataset is, but that should go in a separate PR

On the pydantic side of the codebase, we can implement this with a function that tries validation of the different flavors of zarr group, from most specific to least specific, and returns the first class that matches?

src/ome_zarr_models/v04/bioformats2raw.py
```diff
@@ -0,0 +1,16 @@
...
+
+from pydantic import Field
+
+from ome_zarr_models.base import Base
+from ome_zarr_models.v04.plate import Plate
+
+
+class BioFormats2Raw(Base):
+    """
+    A bioformats2raw zarr group.
```
d-v-b, line 11: does this model a zarr group, or the attributes inside that group?

dstansby: Just the attributes inside that group - should I rename to `BioFormats2RawAttributes` then?

d-v-b: maybe `BioFormats2RawAttrs`, in keeping with our use of the same abbreviation over in multiscales.py

d-v-b (approved):
lgtm!

PR #52, opened 2024-11-23
Clean up multiscales

dstansby:
Some work towards https://github.com/BioImageTools/ome-zarr-models-py/issues/33.

I've tried to clean this up so that `multiscales.py` just contains the multiscales data model. Even if that part of the spec contains some validations we need zarr arrays for, for clarity I think it's best to do these elsewhere. I then relaised that my `Image` class was intended to be the same as @d-v-b's `MultiscaleGroup` group, so I deleted `Image` and renamed `MultiscaleGroup` to image. I think this makes sense, but I'm a bit unsure so folks should definitely feel free to disagree/suggest alternatives.

There's a few minor additions of types where they're used > once too.


d-v-b:
+1 for the renaming, `MultiscaleGroup` doesn't exactly roll of the tongue like `Image`

d-v-b (approved):
looks good!

PR #59, opened 2024-11-27
Update tutorial to demonstrate image reading

dstansby:
This:

- Adds a bit of sample data we can use with the tutorial (and perhaps tests later?)
- Demonstrates how to read this data in the tutorial, and shows a plot
- Modifies Image a bit to hide away pydantic_zarr from the public API. Instead of sub-classing this just uses GroupSpec to do validation and then store the attributes, but these are provided to the user in the form of custom properties.

I'm not 100% sure this is the best way to go in terms of what our "top level" object is and looks like, but I'm wary about sub-classing from another library as we'll lose the flexibility of doing what we want. Very open to other suggestions/ideas though, ideally with modifications to the tutorial to show how they would work as public API.


docs/tutorial.py
```diff
@@ -50,8 +57,12 @@
 # Although these models do not handle reading or writing data, they do expose the zarr
 # arrays.
 
-zarr_arr = ome_zarr_image.group[multiscales_meta[0].datasets[0].path]
-pprint(zarr_arr)
+print(ome_zarr_image.arrays)
+full_res_path = ome_zarr_image.multiscales[0].datasets[0].path
+zarr_arr = ome_zarr_image.arrays[full_res_path]
```
d-v-b, line 62: I'm trying to think of ways to improve this, so that users could do this in one line instead of 2. Given that the only information in the raw `datasets` metadata is "what's the array called, and what are its transforms", the first thing that comes to my mind is to model `datasets` as `Mapping[str, tuple[zarr.Array, ValidTransforms]]`, or `tuple[tuple[zarr.Array, ValidTransforms], ...]`. 

I'm guessing almost every time a user accesses the `path` attribute, it will be immediately prior to getting a reference to the zarr array at that path, so we can shortcut that step by putting references to the zarr arrays there already. The `Mapping` option puts the `path` attribute in the key space of the mapping, but you can also always get the path from the `path` attribute of the zarr arrays, and if that's considered fine than maybe the tuple form is simpler (and ordered). Thoughts?

d-v-b: (not a blocker for this PR)

dstansby: How about adding an `.array` property to `Dataset`, that returns a `zarr.Array` object at the path stored in that dataset? So you could do this instead:

```python
zarr_arr = full_res_path = ome_zarr_image.multiscales[0].datasets[0].array
```

dstansby: I thought more about this, and I don't think what I suggested above is possible, because it would require a reference to the zarr Group, which `Dataset` doesn't have.

src/ome_zarr_models/v04/image.py
```diff
@@ -79,15 +79,31 @@ class ImageAttrs(Base):
...
     """
 
-    _check_arrays_compatible = model_validator(mode="after")(_check_arrays_compatible)
+    def __init__(self, group: zarr.Group) -> None:
+        self._spec = self._get_spec(group)
+        self._group = group
 
     @classmethod
-    def from_zarr(cls, node: zarr.Group) -> Self:
+    def _from_spec(cls, spec: _ImageSpec) -> Self:
```
dstansby, line 101: This is a bit of a hack for testing

dstansby:
I need to fix the tutorial, but would be good to hear your thoughts on this @d-v-b, in particular:

> Modifies Image a bit to hide away pydantic_zarr from the public API. Instead of sub-classing this just uses GroupSpec to do validation and then store the attributes, but these are provided to the user in the form of custom properties.

dstansby:
There's a better version of this at https://github.com/BioImageTools/ome-zarr-models-py/pull/64

PR #60, opened 2024-11-29
Clean up transformations a bit

dstansby:
This cleans up the Transformations objects a bit so that there's now a single `Scale` and `Translation` class, and the `scale` or `translation` key in each of them can take either a path or a vector. I think this makes everything a bit clearer by not having lots of different classes, and makes checking for `ndim` a bit cleaner too.

This also finishes off the coordinateTransformations part of the spec, so fixes https://github.com/BioImageTools/ome-zarr-models-py/issues/32.

### DIFF
--- a/docs/tutorial.py
+++ b/docs/tutorial.py
@@ -6,7 +6,7 @@ from rich.pretty import pprint
 
 from ome_zarr_models.v04 import Image
 from ome_zarr_models.v04.coordinate_transformations import (
-    VectorTranslation,
+    Translation,
 )
 
 # ## Creating models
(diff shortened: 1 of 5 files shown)


d-v-b:
How would we deserialize a transform like `{"type": "scale", "path": "foo/bar"}` after these changes? I feel like without a class that models this JSON exactly (e.g., the `PathScale` classes being removed here) that becomes unwieldy?

d-v-b:
to be clear, I'm all in favor of simplifying things as much as we can, but unfortunately from a structural point of view (i.e., distinguishing types based on their structure) the spec defines two distinct scale transforms, and two distinct translation transforms, and to model that closely we end up needing separate classes for each flavor.

dstansby:
🤦 of course, this does not follow the spec - thanks for pointing that out! Sorry for the noise...

PR #63, opened 2024-12-01
Improve image-labels metadata

dstansby:
Work towards https://github.com/BioImageTools/ome-zarr-models-py/issues/36. Not quite finished, but I put everything still left to do in a TODO: comment in the code so it should be easy to tick them off later.

### DIFF
--- a/src/ome_zarr_models/v04/image.py
+++ b/src/ome_zarr_models/v04/image.py
@@ -72,13 +72,11 @@ class ImageAttrs(Base):
         min_length=1,
     )
     omero: Omero | None = None
-    image_labels: Annotated[
-        ImageLabel | None, Field(..., serialization_alias="image-label")
-    ] = None
+    image_labels: Annotated[ImageLabel | None, Field(..., alias="image-label")] = None
 
     # TODO: validate:
-    # "image-label groups MUST also contain multiscales metadata and the two "datasets" series
-    #  MUST have the same number of entries."
+    # "image-label groups MUST also contain multiscales metadata and the two "datasets"
+    #  series MUST have the same number of entries."
 
 
 class Image(GroupSpec[ImageAttrs, ArraySpec | GroupSpec]):
--- a/src/ome_zarr_models/v04/image.py
+++ b/src/ome_zarr_models/v04/image.py
@@ -76,7 +76,9 @@ class ImageAttrs(Base):
         ImageLabel | None, Field(..., serialization_alias="image-label")
     ]
 
-    # TODO: validate: if image_labels is present, multiscales must be present too
+    # TODO: validate:
+    # "image-label groups MUST also contain multiscales metadata and the two "datasets" series
+    #  MUST have the same number of entries."
 
 
 class Image(GroupSpec[ImageAttrs, ArraySpec | GroupSpec]):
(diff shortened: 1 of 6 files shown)


src/ome_zarr_models/v04/image.py
```diff
@@ -71,6 +72,11 @@ class ImageAttrs(Base):
         min_length=1,
     )
     omero: Omero | None = None
+    image_labels: Annotated[ImageLabel | None, Field(..., alias="image-label")] = None
```
d-v-b, line 75: are we allowed by the spec to write this to JSON as `null` if the field is unset?

src/ome_zarr_models/v04/image_label.py
```diff
@@ -46,23 +49,24 @@ class Source(Base):
     Source data for the labels.
     """
 
-    # TODO: add validation that this path resolves to something
-    image: str | None = "../../"
+    # TODO: add validation that this path resolves to a zarr image group
+    image: str | None = Field(
+        default="../../", description="Relative path to a Zarr group of a key image."
```
d-v-b, line 37: Do we need a default value here?

src/ome_zarr_models/v04/image_label.py
```diff
@@ -46,23 +49,24 @@ class Source(Base):
...
+    label_value: int = Field(..., alias="label-value")
 
 
-def _parse_colors(colors: list[Color] | None) -> list[Color] | None:
+def _parse_colors(colors: tuple[Color] | None) -> tuple[Color] | None:
     if colors is None:
         msg = (
-            "The field `colors` is `None`. Version 0.4 of"
-            "the OME-NGFF spec states that `colors` should be a list of "
+            "The field `colors` is `None`.`colors` should be a list of "
```
d-v-b, line 52:
```suggestion
            "The field `colors` is `None`. `colors` should be a list of "
```

d-v-b:
looks good, I have 2 questions that I left as in-line comments

src/ome_zarr_models/v04/image_label.py
dstansby: 🤷 the spec says
> The default value is "../../" since most labels are stored under a subgroup named "labels/" (see above).

So I guess we should?

d-v-b: yeah i guess so. weird language to have in a spec document but 🤷‍♂️

src/ome_zarr_models/v04/image.py
dstansby: Probably not. We can use the `exclude_unset` keyword argument to `model_dump()` to make sure we're not dumping any unset, but I think that still leaves open the option that someone sets the value to `None` then writes 😢 looking at https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields I'm not sure there's a way round this though - perhaps we should just document it as a known issue?

d-v-b: over in `pydantic-ome-ngff` I used a special attribute to track the fields that shouldn't be serialized if they are null. we might want to do something similar. But until then I would suggest we keep this as-is and put in a TODO statement about dealing with the null value

dstansby: 👍 I've put in a "known issue" note on the homepage of the docs, definiltey up for us fixing this in a follow up though

PR #64, opened 2024-12-02
Add a simpler image class

dstansby:
This adds a simpler image class, that stores a single group that represents an OME-zarr image. The attributes are taken from this group, and validated. Alongside this I've updated the tutorial to show how to:

- Create an `Image2` object
- Read, modify, and then write back some metadata

I think the new image class is a lot simpler than the existing one that depends on pydantic_zarr, and conceptually easier to understand - an OME-zarr image *is* a zarr group, so the single attribute is a `zarr.Group`.

In terms of functionality for updating metadata and data at the same time so it remains self consistent, I'm imagining we would write a separate submodule (or methods on `Image`) that could do these operations. Things like "add another multiscales level for me".

@lorenzocerrone @d-v-b what do you think about this? If you have positive thoughts, then I can go ahead and replace the exsiting `Image` class with this new `Image2` version, and update the tests accordingly.


dstansby:
As part of my mini-sprint to get a proof of concept working, I'm going to merge this. Functionally this removes the dependency on `pydantic-zarr`, which I think makes this package simpler and easier to implement/understand for now. I am definitely open to putting this back in the future if there's a need, but for now I think we can roll along without it. Sorry if I am missing something here...

d-v-b:
> I think the new image class is a lot simpler than the existing one that depends on pydantic_zarr, and conceptually easier to understand - an OME-zarr image _is_ a zarr group, so the single attribute is a `zarr.Group`.

I'm not sure I agree that this is easier to understand. given that this image class has a reference to a zarr group, which can change, so now this class has to handle bi-directional synchronization between itself and the underlying zarr data. The JSON serialization of this class is also now challenging, because there's no way to serialize a zarr group to JSON. 

The pydantic-zarr-based design does away with both of these problems, so it might be good to clarify what exactly was confusing about it, and maybe we can find a way to fix that.

PR #95, opened 2024-12-18
Clean up coordinate transforms

dstansby:
This improves validation messages and code structure by:

- Having Scale/Translation objects that have both vector and path fields, but validate that only one of them is not None.
- Improving the error message when validating the type of transform

Fixes https://github.com/BioImageTools/ome-zarr-models-py/issues/94


d-v-b:
IMO this is not the most elegant solution to that error message. If we go by the structure of the data defined in the spec, there are 4 types of transforms, not 2. I think a better solution to that error message is to write our own error message, instead of deviating from the types defined in the spec (even though they are very clunky types)

dstansby:
Hmm, I dunno - https://ngff.openmicroscopy.org/0.4/index.html#trafo-md doesn't seem to explicitly define 4 data structures, just says that on the data structures either path or <vector field name> can be defined.

dstansby:
Okay, I unded my over-zealous refactoring, and this is a minimal change needed to make the error message nicer now

d-v-b:
> Hmm, I dunno - https://ngff.openmicroscopy.org/0.4/index.html#trafo-md doesn't seem to explicitly define 4 data structures, just says that on the data structures either path or "name of vector" can be defined.

In python we can't really model a type with an optional (i.e. potentially missing) attribute, only an attribute might be `None`. Put another way, Python types require "optionality" to be expressed as a kind of _value_, not in the structure of the type itself. The python type system also doesn't really let us express types defined by statements like "if such and such a field is `None`, then this other field must not be `None`". So with that in mind, if we want to _closely_ model the transforms defined in the spec as types, we basically have write out 4 separate types. In a type system that allowed dict-like types with truly missing fields _and_ fields that exist conditioned on the values of other fields, then maybe we could do it with just 2 types, but we don't have a language like that :)

d-v-b:
fwiw I tried using just 2 types in `pydantic-ome-ngff`, but I ended up writing a lot of input validation and special serialization logic, and the actual data structures were confusing because they had a combination of fields (`path` and `translation`) that could never appear in actual metadata. So I just accepted that the spec does define 4 types, even if I would rather it only defined 2.

d-v-b (approved):

dstansby:
That all makes sense - thanks for taking the time to explain!

PR #96, opened 2024-12-18
Add a labels property to Image

dstansby:
This adds a new type of dataset, `Labels`, and logic for getting the labels from an `Image` (if they exist).

cc @will-moore - does this look good for the funcitonality you were after?


src/ome_zarr_models/v04/image.py
```diff
@@ -92,7 +93,7 @@ def from_zarr(cls, group: zarr.Group) -> Self:
             A Zarr group that has valid OME-NGFF image metadata.
         """
         # on unlistable storage backends, the members of this group will be {}
-        guess = GroupSpec.from_zarr(group, depth=0)
+        guess = GroupSpec.from_zarr(group, depth=1)
```
dstansby, line 96:
```suggestion
        guess = GroupSpec.from_zarr(group, depth=0)
```

Issue #26, opened 2024-11-22
High level API to find the ngff version from a Group


lorenzocerrone:
The idea is to have a function that tries to find the ngff version from a given group.

```python
def find_version(group: Group) -> str
```

Should this also validate the entire model? Or should it try just to load the `version` attr in the multiscale group?

PR #105, opened 2025-01-05
Add function to load any OME-Zarr dataset

dstansby:
Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/26. Would be good to get feedback on naming for new classes and functions I've added here!

### DIFF
--- a/src/ome_zarr_models/__init__.py
+++ b/src/ome_zarr_models/__init__.py
@@ -1,6 +1,7 @@
 from importlib.metadata import PackageNotFoundError, version
 
 import zarr
+from pydantic import ValidationError
 
 import ome_zarr_models.v04.hcs
 import ome_zarr_models.v04.image
@@ -37,7 +38,7 @@ def load_ome_zarr_group(group: zarr.Group) -> BaseGroup:
     for group_cls in _V04_groups:
         try:
             return group_cls.from_zarr(group)
-        except Exception:
+        except ValidationError:
             continue
 
     raise RuntimeError(f"Could not find any matches for group {group}")
--- a/src/ome_zarr_models/__init__.py
+++ b/src/ome_zarr_models/__init__.py
@@ -1,6 +1,43 @@
 from importlib.metadata import PackageNotFoundError, version
 
+import zarr
+
+import ome_zarr_models.v04.hcs
+import ome_zarr_models.v04.image
+import ome_zarr_models.v04.image_label
+import ome_zarr_models.v04.labels
+import ome_zarr_models.v04.well
+from ome_zarr_models.base import BaseGroup
+from ome_zarr_models.v04.base import BaseGroupv04
+
 try:
     __version__ = version("ome_zarr_models")
 except PackageNotFoundError:  # pragma: no cover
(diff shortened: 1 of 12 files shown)


src/ome_zarr_models/__init__.py
```diff
@@ -1,6 +1,43 @@
...
+
+    This function will 'guess' which type of OME-Zarr data exists by
+    trying to validate each group metadata definition against your data.
+    If validation is successful, that data class is returned without
+    trying any more.
+    """
+    for group_cls in _V04_groups:
+        try:
+            return group_cls.from_zarr(group)
+        except Exception:
```
d-v-b, line 40: maybe we should only catch pydantic validation errors? otherwise informative zarr errors will also get swallowed up here

src/ome_zarr_models/__init__.py
```diff
@@ -1,6 +1,43 @@
...
+def load_ome_zarr_group(group: zarr.Group) -> BaseGroup:
+    """
+    Create an ome-zarr-models object from an existing OME-Zarr group.
+
+    This function will 'guess' which type of OME-Zarr data exists by
+    trying to validate each group metadata definition against your data.
+    If validation is successful, that data class is returned without
+    trying any more.
+    """
+    for group_cls in _V04_groups:
```
d-v-b, line 37: is `_V04_groups` ordered from most-to-least specific? Or is it otherwise impossible for a given zarr group to match two members of `_V04_groups`?

src/ome_zarr_models/__init__.py
```diff
@@ -1,6 +1,43 @@
...
+_V04_groups: list[type[BaseGroupv04]] = [
+    ome_zarr_models.v04.hcs.HCS,
+    ome_zarr_models.v04.image_label.ImageLabel,
+    ome_zarr_models.v04.image.Image,
+    ome_zarr_models.v04.labels.Labels,
+    ome_zarr_models.v04.well.Well,
+]
+
+
+def load_ome_zarr_group(group: zarr.Group) -> BaseGroup:
```
d-v-b, line 28: I think we can indulge in a shorter function name -- how about `read`

src/ome_zarr_models/__init__.py
dstansby: This is just alphabetical, but any order is fine because every group has at least one required field, and none of the field names are shared between different groups.

src/ome_zarr_models/__init__.py
dstansby: 👍 for shorter name - perhaps `open()` to copy the top level methods in `zarr-python`?

dstansby: I went with `open_ome_zarr` (since `open` is a builtin)

d-v-b: `open` is a bit ambiguous about what comes out (is the result mutable? does it require a file handle?) which is why I think `open` was not a great choice for `zarr-python`. If the thing returned by this function is immutable, then i'd pick a name that emphasizes this fact

PR #111, opened 2025-01-08
Add typos pre-commit hook, and fix typos

imagejan:


d-v-b (approved):

dstansby:
Huh, I'm surprised that the codespell hook (which should also check spelling) didn't catch these. This is definitely an improvement, thanks for the contribution!

PR #126, opened 2025-01-23
Add test data from ngff

LucaMarconato:
- Move test data in a subfolder to distinguish the data from the NGFF specs and new test data that we will create
- Add README for test data to keep track of which data we are using so far, and to give an order on what we will test first


d-v-b:
quick q: since all of the examples are already on github, can't we refer to them by URL instead of copying them here as files?

LucaMarconato:
Sorry maintainers, I am still not familiar with the GitHub APIs from my IDEs. I wanted to open this against our fork not directly here; I will close the PR.

LucaMarconato:
> quick q: since all of the examples are already on github, can't we refer to them by URL instead of copying them here as files?

Good point, the answer is in the readme that I added:

> Note: Some of the data from the specification needs fixes. Once the project is in the late stage, it will be important to diff the data with the one from the specification and make a PR to fix the issues in the NGFF specs.

We saw some things to fix, and it will be easier to fix it locally and then make a final PR to NGFF to address them.

Issue #139, opened 2025-01-29
Clarify release cadence and versioning policy


ziw-liu:
From a downstream developer's perspective, it would be great if the documentation contains information about the release cadence, as well as versioning policy, i.e. semantic versioning (or not) and its interpretations for the 0.x stage.

PR #142, opened 2025-01-30
Add a versioning scheme

dstansby:
@ziw-liu is there anything I missed here that would be helpful for you?

Should fix https://github.com/ome-zarr-models/ome-zarr-models-py/issues/139

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -55,7 +55,7 @@ _Note:_ support for OME-Zarr 0.5 is not complete, but when it is the following i
 
 ## Versioning
 
-`ome-zarr-models` has a 0.minor.micro versioning scheme where:
+`ome-zarr-models` has a major.minor versioning scheme where:
 
 - The major version is incremented when support for a new version of the OME-Zarr specification is added, or a breaking change is made to the package.
 - The minor version is incremented for any other changes (e.g., documentation improvments, bug fixes, new features)
--- a/docs/index.md
+++ b/docs/index.md
@@ -58,7 +58,7 @@ _Note:_ support for OME-Zarr 0.5 is not complete, but when it is the following i
 `ome-zarr-models` has a major.minor versioning scheme where:
 
 - The major version is incremented when support for a new version of the OME-Zarr specification is added, or a breaking change is made to the package.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -44,6 +44,17 @@ Developers of this package are active on our [Zulip chat channel](https://images
   This is because it is transitional, and we have decided to put time into implementing other
   parts of the specification. We would welcome a pull request to add this functionality though!
 
+## Versioning
+
+`ome-zarr-models` has a 0.minor.micro versioning scheme where:
```
will-moore, line 49: You have `minor.micro` here but describe `major` and `minor` below?

dstansby: 🤦 wooops, thanks for catching!

dstansby, line 49:
```suggestion
`ome-zarr-models` has a major.minorversioning scheme where:
```

dstansby:
pre-commit.ci autofix

ziw-liu:
Thanks @dstansby! Do I understand correctly that, after 1.0, this package will _not_ follow semantic versioning, as indicated by the lack of a 'patch' number, and each bug fix will increment the 'minor' version?

dstansby:
Yes, I don't think we're going to follow semantic versioning. Will that cause any problems for you/do you think it would be better if we did?

ziw-liu:
> Yes, I don't think we're going to follow semantic versioning. Will that cause any problems for you/do you think it would be better if we did?

Mainly for separating feature updates from 'just bugfixes' patches. The [spec](https://semver.org/) has the full explanation on this.

dstansby:
For downstream users, my thoughts were that from a practical perspective it doesn't really matter if a new release contains bug fixes/new features/better docs/other stuff, as long as it's backwards compatible.

and from a maintenance perspective, it's easier to have two (major.minor) version numbers instead of three. I think I see the benefits of semantic versioning if a project is going to support multiple minor versions in parallel (e.g., 1.2, 1.3, 1.4 - [Matplotlib](https://matplotlib.org/) is an example of a project that does this), but since we're not going to do this, I think we should stick with major.minor.

Unless there's anything I've missed/any other thoughts, I'll merge this in a few days (but comments welcome after merging too!)

Issue #145, opened 2025-02-04
Fix version number in v05 multiscales


dstansby:
> Hi @dstansby - I was just testing a tool to generate `TypeScript` types from Pydantic - see https://github.com/marketplace/actions/pydantic-to-typescript
> 
> When I ran it on the _v05 classes...
> 
> ```
> $ pydantic2ts --module src/ome_zarr_models/_v05/image.py --output image05.ts
> ```
> 
> It gave me a `Multiscale` with `version?: "0.4" | null;` and I also don't see the `"version": "0.5"` which is required on `ImageAttrs` I think?
> 
> Apologies if this is already on your radar but I didn't recognise it in the issues above. Thanks. 
> 
> ```
> /**
>  * Model for the metadata of OME-Zarr data.
>  *
>  * See https://ngff.openmicroscopy.org/0.4/#image-layout.
>  */
> export interface ImageAttrs {
>   /**
>    * The multiscale datasets for this image
>    *
>    * @minItems 1
>    */
>   multiscales: [Multiscale, ...Multiscale[]];
>   omero?: Omero | null;
>   [k: string]: unknown;
> }
> /**
>  * An element of multiscales metadata.
>  */
> export interface Multiscale {
>   axes: Axis[];
>   /**
>    * @minItems 1
>    */
>   datasets: [Dataset, ...Dataset[]];
>   version?: "0.4" | null;
>   coordinateTransformations?: [unknown] | [unknown, unknown] | null;
>   metadata?: {
>     [k: string]: unknown;
>   };
>   name?: JsonValue | null;
>   type?: {
>     [k: string]: unknown;
>   };
>   [k: string]: unknown;
> }
> ```
>  

 _Originally posted by @will-moore in [#88](https://github.com/ome-zarr-models/ome-zarr-models-py/issues/88#issuecomment-2634469500)_

dstansby:
I've removed the version number in v05 over in #146 (thanks for reporting, this was not on our list and should have been!). For

> and I also don't see the "version": "0.5" which is required on ImageAttrs I think?

`ImageAttrs` models the metdata that sits under the 'ome' key, so shouldn't contain the ome version, as that also sits directly under the 'ome' key. Does that make sense?

will-moore:
From https://ngff.openmicroscopy.org/0.5/index.html#multiscale-md we get:

```
"attributes": {
    "ome": {
      "version": "0.5",
      "multiscales": [
        {
...
```
[...]

PR #149, opened 2025-02-04
Fix location of OME version in 0.5

dstansby:
See https://github.com/ome-zarr-models/ome-zarr-models-py/issues/145#issuecomment-2634669241.


will-moore:
AFAICT this looks good. Certainly the `version` is in the right place alongside `multiscales` for `ImageAttrs` 👍

PR #154, opened 2025-02-05
Explain which 0.5 spec version we validate against

dstansby:


docs/index.md
```diff
@@ -46,6 +46,9 @@ Developers of this package are active on our [Zulip chat channel](https://images
 
 ### OME-Zarr 0.5
 
+- Since the first release of OME-Zarr 0.5 (commit [8a0f886](https://github.com/ome/ngff/tree/8a0f886aac791060e329874b624126d3530c2b6f)), the specification has edited without the version number in OME-Zarr datasets being changed.
+  A diff between the 'current' 0.5 specification and 'original' 0.5 specification [can be seen here](https://github.com/ome/ngff/compare/0.5.0...main#diff-6e0c0575683d2ac5c07564e6828e9c71ae3b93b6eacc36575055150af6c5ef73).
```
will-moore, line 50: I tried to find a way to link to just the `0.5/index.bs` file within that diff, since that's the only one we care about? But couldn't see that github supports that.

d-v-b (approved):

PR #165, opened 2025-02-10
Finish refactoring validators

dstansby:
This finishes a refactoring of validators to make them all methods on classes. This:

- means where validators are defined consistent (previously, we had a mix of methods and functions)
- makes the field listing a bit easier to read, by removing lots of `Annotated` definitions that include validators. Now there's just the data type defined, which I think makes it more readable

### DIFF
--- a/src/ome_zarr_models/common/multiscales.py
+++ b/src/ome_zarr_models/common/multiscales.py
@@ -101,19 +101,6 @@ def _ensure_scale_translation(
     return transforms_obj
 
 
-def _ensure_axis_length(axes: Axes) -> Axes:
-    """
-    Ensures that there are between 2 and 5 axes (inclusive)
-    """
-    if (len_axes := len(axes)) not in VALID_NDIM:
-        msg = (
-            f"Incorrect number of axes provided ({len_axes}). "
-            "Only 2, 3, 4, or 5 axes are allowed."
-        )
-        raise ValueError(msg)
-    return axes
-
-
 class Dataset(BaseAttrs):
     """
     An element of Multiscale.datasets.
@@ -149,10 +136,7 @@ class MultiscaleBase(BaseAttrs):
     An element of multiscales metadata.
     """
 
-    axes: Annotated[
-        Axes,
-        AfterValidator(_ensure_axis_length),
-    ]
+    axes: Axes
     datasets: tuple[Dataset, ...] = Field(..., min_length=1)
     coordinateTransformations: ValidTransform | None = None
     metadata: JsonValue = None
@@ -242,6 +226,20 @@ class MultiscaleBase(BaseAttrs):
                 )
 
         return datasets
+
+    @field_validator("axes", mode="after")
+    @classmethod
+    def _ensure_axis_length(cls, axes: Axes) -> Axes:
+        """
+        Ensures that there are between 2 and 5 axes (inclusive)
+        """
+        if (len_axes := len(axes)) not in VALID_NDIM:
+            msg = (
+                f"Incorrect number of axes provided ({len_axes}). "
+                "Only 2, 3, 4, or 5 axes are allowed."
+            )
+            raise ValueError(msg)
+        return axes
 
     @field_validator("axes", mode="after")
     @classmethod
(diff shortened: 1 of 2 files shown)


src/ome_zarr_models/common/multiscales.py
```diff
@@ -207,43 +59,83 @@ def build(
...
-        s1, s2 = scales[i], scales[i + 1]
-        is_ordered = all(s1[j] <= s2[j] for j in range(len(s1)))
-        if not is_ordered:
-            raise ValueError(
-                f"Dataset {i} has a lower resolution (scales = {s1}) "
-                f"than dataset {i+1} (scales = {s2})."
+        class Transforms(BaseModel):
+            transforms: list[Transform]
+
+        transforms = Transforms(transforms=transforms_obj).transforms
```
d-v-b, line 82: this looks odd, what is this `Transforms` class for exactly?

dstansby: Reading the comment above, this is to parse the input as a list of `Transform`, which we need to do because this is a "before" validator. I'm not 100% why we don't just have this as an after validator so we don't need to bother with this though...

d-v-b:
Generally i'm not a fan of moving all of these functions to class methods. Functions are reusable (and can be tested in isolation), while class methods generally are not. And methods can always be made to wrap the functions, but the reverse is not true.

A lot of these validation routines are narrowly scoped -- ensuring that all the transforms in a sequence have consistent dimensionality only requires a sequence of transforms, not an entire multiscales object. Making that validation routine a method doesn't help that validation routine, but it does hurt anyone who wants to use the validation function on its own, outside of the context of object creation.

If the main goal is ensuring that every validator is a class method, and removing all the `Annotated` stuff, then there's no need to remove functions that are already defined. Just wrap them as classmethods.

dstansby:
>  Functions are reusable (and can be tested in isolation), while class methods generally are not.

It's easy enough to call classmethods, the call signature is just `MyClass.my_method(...)` instead of `my_function(...)`. Definitely appreciate this might be more confusing to users, but none of our validation code is user-facing API, so my main thought when re-organising this code was what makes the code itself easier to read for developers. 

In terms of making it easier to read/understand, my thoughts were it's clearer if the validation code lives in the same place as the data it's validating (ie, on the class). Most of the validation code isn't used in multiple places, so I think it's worth sacrificing re-usability for ease-of-understanding. In other words, "all the validation for this class is defined below on this class" is simpler than "all the validation for this class is in all the different places that these functions are defined".

where validation code is needed multiple times, and is generic enough that it's API doesn't reference a particular class I started [`ome_zarr_models.common.validation`](https://github.com/ome-zarr-models/ome-zarr-models-py/blob/main/src/ome_zarr_models/common/validation.py).

dstansby:
So in summary, I agree in general with https://github.com/ome-zarr-models/ome-zarr-models-py/pull/165#issuecomment-2647785875 *if* code is intended to be used more than once, or is useful to users. But I think the code moved in this PR doesn't fall into either of those two categories, so for ease of understanding code is better as classmethods. Does that make sense?

d-v-b:
> > It's easy enough to call classmethods, the call signature is just `MyClass.my_method(...)` instead of `my_function(...)`.
> 
> "easy to call" isn't the same thing as "reusable". If I have unrelated 2 classes that need the same validation logic, and the vaidation is implemented as a function, then both classes can use that function. If instead the validation logic is implemented as a class method on class A, then class B has to have a reference to class A, and the functions aren't unrelated any more.
> 
> > In other words, "all the validation for this class is defined below on this class" is simpler than "all the validation for this class is in all the different places that these functions are defined".
> 
> I don't think "all the validation for this class is defined in the same module as this class" is so bad. But if this PR is a response to a specific complaint from a user / developer, then feel free to ignore

> if code is intended to be used more than once,

It's not in general possible to plan to use code in exactly N places, so I think a good strategy is make things re-usable by default.

d-v-b:
> Are there methods moved that you think will in the future be shared across two unrelated classes?

I can't think of anything at the moment, and this would be easily handled in the future because none of these functions are very complex or long. So this is much more a style concern than anything else.

> I think it's much nicer to have classes first, which are the context for the validation.

Personally I wouldn't let the aesthetics of the source code document determine its logical structure. But the stakes here are pretty low, so I'm happy to go along with these changes.

dstansby:
I'm going to merge this so at least we are consistent in where validation code lives (even if it doesn't turn out to be the best option), and I can take a shot at implementing RFC-3.

d-v-b:
sounds good!

Issue #143, opened 2025-01-30
Document (and decide?) how to write metadata


dstansby:
We should decide how users should write out metdata, and document this. Previous discsusion on the topic can be found at https://github.com/ome-zarr-models/ome-zarr-models-py/issues/121#issuecomment-2598363436, but I thought I'd split this out to make it more visible.

dstansby:
I started scratching my head about this, and my initial thought was the model for creating metadata could be:

1. Create a in-memory group (ie `pydantic_zarr.GroupSpec`) without any datasets. Something like `Image.new(...)`
2. Add datasets one-by-one using a method that would also automatically add the relevant `ArraySpec` to the in-memory group. Something like `new_image.add_dataset(path, transforms, array)`
3. Save to storage

I ran into the problem that at least in `Image`, it's not valid to have zero multiscale datasets defined at step 1. Perhaps this is something that could be relaxed in the spec in the future (to allow for zero datasets?)

Does anyone have any other ideas here? I feel like @d-v-b had

d-v-b:
to use ome-zarr metadata you need a collection of named arrays (a multiscale pyramid) with some spatial information for each array. Without this, you can't start creating any meaningful metadata. So I think we should assume that users have this already, and we should provide a top-level function that takes some arrays, some names, some spatial transformation info, and a bunch of other keyword arguments (axis names, axis types, etc), and creates a `GroupSpec` instance. There should probably be 2 functions: one that just returns the `GroupSpec` instance, for people who want to actually create their zarr group later, and a function that takes additional `store` and `path` keyword arguments and creates a Zarr group, and returns a handle to the group. Users can then save their arrays as they see fit.

I'm could imagine a situation where people don't know ahead of time how many arrays they will have, or what their shapes are, etc, but I think this is super unlikely. We should probably assume that people have already decided how they are downsampling, etc, and take for granted that all the arrays are defined already.

PR #167, opened 2025-02-15
Proof of concept for creating/writing Image metadata

dstansby:
This is a first pass at a helper to create and writing metadata. It's based of @d-v-b's code in the tests, with a couple of changes:

- The input arrays are `zarr.Array` instead of numpy arrays + zarr metdata. This cleans up the function signature, and avoids us having to re-implement the interface for creating zarr ararys here.
- Added parameters for specifying global translation/scale transformations.

Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/143


docs/tutorial.py
```diff
@@ -53,3 +57,41 @@
...
+
+# ## Creating new datasets
+#
+# To create new OME-Zarr datasets, the ``.new()`` method on the OME-Zarr groups
+# can be used.
+#
+# As an example we'll create an OME-Zarr image with two arrays, one at the
+# original resolution and one downsampled version.
+
+arrays = [zarr.empty(shape=(100, 100)), zarr.empty(shape=(50, 50))]
```
d-v-b, line 69: So for users who have data represented as a numpy or dask array, they have to first create in-memory zarr arrays before they can create on-disk zarr arrays? This seems needlessly indirect?

docs/tutorial.py
```diff
@@ -53,3 +57,41 @@
...
+pixel_unit = "um"
+
+ome_zarr_image = Image.new(
+    arrays=arrays,
+    paths=["level0", "level1"],
+    axes=[
+        Axis(name="y", type="space", unit=pixel_unit),
+        Axis(name="x", type="space", unit=pixel_unit),
+    ],
+    global_scale=pixel_size,
```
d-v-b, line 80: I would not include the global scale here, because it's not the strongest part of the spec and it doesn't add anything to the data model

dstansby: I use this all the time to specify the pixel size of my lowest resolution voxels, which then makes the scale transforms of each level nicer because they are then (1, 1), (2, 2), (4, 4).

dstansby: I also think we should make sure the function is flexible enough for users to be able to specify all possible fields in one shot, instead of calling this function and then having to modify metdata afterwards.

d-v-b: It might make them nice integers but it also makes them inconsistent with the units defined in the `Axis` metadata. It also means the spec allows the same thing (the coordinates of an image) to be defined in two different ways, which is a bug IMO

d-v-b: e.g., if I'm looking at some image metadata, it really doesn't help me to know that image X has a scale that's 2 times some other number defined elsewhere in the metadata document! I just want to know the actual scale for that image. And similarly for translation.

d-v-b: > I also think we should make sure the function is flexible enough for users to be able to specify all possible fields in one shot, instead of calling this function and then having to modify metdata afterwards.

I'm fine with the global scale as a keyword argument to the method, because this method should support the full metadata model. but for this example in the docs, I would keep things simple and omit it, since it's redundant. Also, I think the global transform stuff is going away in future versions of the spec.

dstansby: I think that is a discussion for the specification, not for this package? My take is that since we're strictly following the spec, if the spec allows something then we should allow users to write it.

FWIW I don't disagree with your reasoning and would support changing the spec to remove the global scale/translation (but maybe that's going to happen anyway in the coordinates RFC?)

d-v-b: > I think that is a discussion for the specification, not for this package? My take is that since we're strictly following the spec, if the spec allows something then we should provide support for users to write it.
> 
> FWIW I don't disagree with your reasoning and would support changing the spec to remove the global scale/translation (but maybe that's going to happen anyway in the coordinates RFC?)

support doesn't mean encourage. for documentation, we should show the core features users need, and not show how to support every single thing in the spec. the global scale IMO is a wart that should not be encouraged, so choosing examples judiciously is actually high-value if it steers people away from anti-patterns.

d-v-b: ofc the global scale should be documented fully in API docs. but how-to guides should probably not show it, unless a bunch of users request this for some reason.

docs/tutorial.py
dstansby: I think the difference would just be where the arguments for specifying the Zarr array config are specified, in `Image.new()` or `zarr.Array` (or whatever the correct function is in `zarr-python` for going from numpy > Zarr arrays).

Two cases why I think we have to accept zarr arrays:

- An array that doesn't fit in memory - it's impossible for me to put my 1TB images into a numpy array, but I can create an array with the right shape/dtype/zarr config with `zarr.empty`, write metdata, then fill up the Zarr arrays bit by bit.
- Someone has a zarr array that's not backed by a numpy array (less sure here, in zarr-python 3 it could be a GPU array, but I'm not too sure about zarr-python 2?)

docs/tutorial.py
dstansby: > so choosing examples judiciously is actually high-value if it steers people away from anti-patterns.

Ah, my bad 🤦 - I thought this comment was on the function signature, not the example - I agree with removing this from example!

docs/tutorial.py
d-v-b: I think accepting zarr arrays is fine, but accepting _only_ zarr arrays seems needlessly restrictive, and requiring that people create an empty, in-memory zarr array just to create some metadata seems a bit indirect. I suspect users will not find it intuitive that they have to create some arrays only to discard them after creating metadata.

For data that doesn't fit in memory or can't be formed into an array-like object, we should ask what representation of that data does the user already have? Presumably they know the shape and the dtype of each array. So there should be a function that takes shape and dtype information, along with the rest of the structure of a zarr array, to define the output Zarr arrays. 

But for users who already have an array-like object (including a zarr array), it's expedient if there's a different function that can take that object and infer the zarr array properties from it. 

That's why I defined two functions: `from_arrays` and `from_array_props`, neither of which has a great name. The first makes metadata from existing array-like objects, and the second makes metadata from properties of array-like objects (for when someone knows the shape and dtype but doesn't have an actual array). And while these are useful as class methods, users should probably be directed to use functions instead of methods, because a function can take a `ome_zarr_version` keyword argument and create different types of metadata objects, but a class method will be tied to a specific version, and so the user-facing API would be less stable.

docs/tutorial.py
dstansby: Global scale removed 👍

d-v-b: thank you!

docs/tutorial.py
dstansby: > requiring that people create an empty, in-memory zarr array just to create some metadata seems a bit indirect. I suspect users will not find it intuitive that they have to create some arrays only to discard them after creating metadata.

👍 

I've had a go at using `pydantic_zarr.v2.ArraySpec` as input instead. In the tutorial, I've given examples of how to construct these from numpy and zarr arrays (thanks to the nice methods in `pydantic_zarr`!). I *think* I quite like this because it signals to users that they're only creating/saving metdata with `ome-zarr-models`, because they're forced to pass and `ArraySpec` instead of anything that contains data. What do you think about that?

> a function can take a ome_zarr_version keyword argument and create different types of metadata objects, but a class method will be tied to a specific version, and so the user-facing API would be less stable.

I'm 👎 to having a function like this, because as the OME-Zarr specification evolves, if we want to support multiple versions in one function signature I think it will end up a mess? Open to reviewing/commenting on a concrete implementation though, perhaps that supporsts OME-Zarr 0.4 *and* RFC-5 coordinate as a test case for what a function that supports a part of the spec that is in flux.

I quite like tying the creation methods to a certain version of OME-Zarr, because it's an easy way for users to understand/choose an OME-Zarr version to write with.

dstansby:
pre-commit.ci autofix

dstansby:
This is good for review now - would be good to get a few pairs of eyes on this and get feedback. @will-moore would you be up for reviewing too, since I know you've thought about writing out metadata in the past?

d-v-b:
> Trying code based on the docs/tutorial.py looks good. The only issue I found so far was the problem of writing `/multiscales/0/name` to `null` which causes a validation fail in ome-ngff-validator since "name must be a string".

Although it's reasonable to assume that the name should be a string, unfortunately the spec doesn't actually give the `name` field a type: https://github.com/ome/ngff/blob/0e62ce8b320c6ce997c93f0b8ba13d89004836d8/0.4/index.bs#L378

dstansby:
Yeah, to echo the above that's a bug with the NGFF validator if it doesn't take into account that `name` (at least in version 0.4) doesn't have to be a string.

From a 'making this package nicer' POV it would definitely be nice to have a way of not writing out `null` values, but I have investigated this and am yet to find a good way... ideas and PRs very welcome though on how to do this 😄

d-v-b:
> From a 'making this package nicer' POV it would definitely be nice to have a way of not writing out `null` values, but I have investigated this and am yet to find a good way... ideas and PRs very welcome though on how to do this 😄

we could copy what I did in `pydantic-ome-ngff` -- use a class attribute to track the names of fields that should not be serialized if their value is `None`, and override `model_dump` accordingly.

d-v-b:
but in the specific case of the `name` field, the ngff validator should accept `null`

will-moore:
This was my test code for writing an image...

```
from skimage import data

fp = "coins.zarr"
store = zarr.DirectoryStore(path=fp, dimension_separator="/")
arr = data.coins()

array_specs = [ArraySpec.from_array(arr)]
pixel_size = (6, 4)
pixel_unit = "um"

ome_zarr_image = Image.new(
    array_specs=array_specs,
    paths=["level0"],
    axes=[
        Axis(name="y", type="space", unit=pixel_unit),
        Axis(name="x", type="space", unit=pixel_unit),
    ],
    scales=[[p * 1 for p in pixel_size]],
    translations=[[0, 0]],
)
ome_zarr_image.to_zarr(store=store, path="/")

# zarr v2
zarr.save(store, arr, path="level0")
```

One thing I tried first was writing the zarr array first and then calling `ome_zarr_image.to_zarr()` to write the metadata but this failed with:

```
zarr.errors.ContainsGroupError: path 'A group already exists at path /. That group is structurally dissimilar to the group you are trying to store.Call to_zarr with overwrite=True to overwrite that group.' contains a group```

I don't know if it's possible to allow that, since I could imagine cases where you might want to write the arrays first?

One other observation / note-to-self (not really related to this PR): When using `open_ome_zarr(zarr_group)` to read an existing image as in the tutorial, it fails if the image is invalid (e.g. missing a 'scale' from coordinateTransformations):
```
RuntimeError: Could not find any matches for group <zarr.hierarchy.Group '/' read-only>
```
whereas `Image.from_zarr(zarr_group)` gives me some useful validation errors (once I'd removed the earlier failing `open_ome_zarr()`).

dstansby:
I think I addressed all the comments, so will merge. I'll do some doc improvements in a follow up PR, and then get this out in a release soon.

PR #169, opened 2025-02-15
Improve group inheritance

dstansby:
This improves the class inheritance of groups, by making the base groups generics. This simplifies creation of the concrete groups.

### DIFF
--- a/src/ome_zarr_models/base.py
+++ b/src/ome_zarr_models/base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Literal
+from typing import Generic, Literal, TypeVar, Union
 
 from pydantic import BaseModel, ConfigDict
 from pydantic_zarr.v2 import ArraySpec, GroupSpec
@@ -23,7 +23,10 @@ class BaseAttrs(BaseModel):
     )
 
 
-class BaseGroup(GroupSpec[BaseAttrs, ArraySpec | GroupSpec], ABC):  # type: ignore[misc]
+T = TypeVar("T", bound=BaseAttrs)
+
+
+class BaseGroup(GroupSpec[T, Union["ArraySpec", "GroupSpec"]], ABC, Generic[T]):  # type: ignore[misc]
     """
     Base class for all OME-Zarr groups.
     """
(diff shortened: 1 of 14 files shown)


src/ome_zarr_models/base.py
```diff
@@ -23,7 +23,10 @@ class BaseAttrs(BaseModel):
     )
 
 
-class BaseGroup(GroupSpec[BaseAttrs, ArraySpec | GroupSpec], ABC):  # type: ignore[misc]
+T = TypeVar("T", bound=BaseAttrs)
+
+
+class BaseGroup(GroupSpec[T, Union["ArraySpec", "GroupSpec"]], ABC, Generic[T]):  # type: ignore[misc]
```
d-v-b, line 29: I'm a bit confused, why does `T` (the type of the attributes) appear twice in the list of superclasses? Since `GroupSpec` inherits from `Generic`, why do we need to inherit from `Generic` again? 

And by hard-coding the the type of the members to `ArraySpec | GroupSpec`, it becomes impossible to define zarr groups that are typed w.r.t. their members -- if a future version of the spec requires that multiscale arrays have specific attributes (which I think is coming with the transforms rfc), then we cannot express that here.

d-v-b: can you describe this "inheritance niceness" metric we are optimizing for 🤔

dstansby: In this PR, changing the definition of the OME-Zarr groups from 
```python
HCS(GroupSpec[BaseZarrAttrs[HCSAttrs], ArraySpec | GroupSpec], BaseGroupv05)
```

to 

```python
class HCS(BaseGroupv05[HCSAttrs])
```

which is simpler because it only inherits from one class, not two as before.

d-v-b: `BaseGroupv05` inherits from `GroupSpec` already, so it's already generic with w.r.t the attributes and members type parameters. 

You can do `class HCS(BaseGroupv05[HCSAttrs, ArraySpec | GroupSpec])`, but you could also use a generic type for the second type parameter:

```python
TMembers = TypeVar('TMembers', bound = ArraySpec | GroupSpec)
class HCS(BaseGroupv05[HCSAttrs, TMembers])
```

dstansby: But then we have to go around putting `TMembers` in the definition of every class, whereas with the current code we don't. Does that give extra flexibility/advantages?

dstansby:
I'll merge this as an improvement on the current state

PR #175, opened 2025-02-26
Improve error message in open_ome_zarr()

dstansby:
See https://github.com/ome-zarr-models/ome-zarr-models-py/pull/167#issuecomment-2682298043. I don't think we have any way to propagate validation errors in this function (because we would have to propagate errors from *all* the OME-Zarr groups), but this improved error message at least explains the route to getting a more precise validation error message.

cc @will-moore

### DIFF
--- a/src/ome_zarr_models/__init__.py
+++ b/src/ome_zarr_models/__init__.py
@@ -52,4 +52,10 @@ def open_ome_zarr(group: zarr.Group) -> BaseGroup[Any]:
         except ValidationError:
             continue
 
-    raise RuntimeError(f"Could not find any matches for group {group}")
+    raise RuntimeError(
+        f"Could not successfully validate {group} with any OME-Zarr group models.\n"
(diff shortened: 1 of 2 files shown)


src/ome_zarr_models/__init__.py
```diff
@@ -52,4 +52,10 @@ def open_ome_zarr(group: zarr.Group) -> BaseGroup[Any]:
         except ValidationError:
             continue
 
-    raise RuntimeError(f"Could not find any matches for group {group}")
+    raise RuntimeError(
+        f"Could not successfully validate {group} with any OME-Zarr group models.\n"
+        "\n"
+        "If you know what type of group you tryint to open, using the "
```
imagejan, line 58:
```suggestion
        "If you know what type of group you are trying to open, using the "
```

will-moore:
👍 Great, thanks.

PR #176, opened 2025-02-28
Add cross linking to Zarr docs

dstansby:
I think linking to the latest v2 docs is the best we can do here? Ideally we would link to whatever the latest v2 release is, but I don't think that's possible...


dstansby:
🤔 this doesn't seem to be working: https://ome-zarr-models-py--176.org.readthedocs.build/en/176/api/#ome_zarr_models.open_ome_zarr

PR #187, opened 2025-05-07
Initial support for RFC5

LucaMarconato:
# Intro
This PR, done together with @melonora, adds initial support for RFC5 by implementing the classes needed for a first example with simple transformations (`Identity`, `Translation`, `Scale`, `Sequence`) but with the transformations already in the within the "full object", i.e. correctly handled in the context of `Dataset`, `Multiscale` and `Image` classes.

This PR also adds a comprehensive testing strategy by adding placeholders (=skipped tests) for all the data-driven tests available for this RFC5, so that the implementation work can be tracked and easily distributed across developers. 

## The tests
There are 3 types of tests.
1) **Tests done only in-memory.** Here small objects are created to quickly cover various edge cases. An example is `test_coordinate_system_name_not_empty()`
2) **Tests from `.json` data living in the repository.** These tests are using the same logic of the tests under `tests/v04/data` and `tests/v05/data`.
3) **Tests against "official" external examples**. Outside these repos there are two sets of examples available. One set of examples [from the specification itself](https://github.com/bogovicj/ngff/tree/coord-transforms/0.6-dev/examples) and another one from [the new repo](https://github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples/tree/main/zarr2) from @bogovicj (the `zarr2` examples are used). The data for these tests is not added to the repository, but made accessible via `git submodule`. (In the beginning we were adding the data to the repo, but after some comments from @d-v-b and @dstansby we changed this).

### Three more comments on point 3)
1. In the examples from the specs, using the whole `ngff` repo inside a submodule was creating problems because the `ngff` repo has tests that would be triggered by living in the `tests` folder of `ome-zarr-models-py`. I therefore use the `sparse-checkout` feature of `git`. We explain this in `tests/_rfc5_transforms/data_rfc5/README.md`.
2. Having dozens of evolving data examples (i.e. new examples may be created) means that it is easy to miss some tests for new data. For this reason we created a test to check that each dataset is tested. This is available in `tests/_rfc5_transforms/test_data_rfc5/test_all_data_is_tested.py`
3. The examples in the `ngff` repo only contain the `coordinateTransformations` part of the metadata, and they are not full examples like the ones in the new repo from John. We handle this by wrapping the metadata of the `ngff` repo inside `Multiscale` objects via the function `wrap_coordinate_transformations_and_systems_into_multiscale()`.

## Comments on the code
1. We added new classes under `src/ome_zarr_models/_rfc5_transforms`. Here we mostly "rebrand" existing `v05` classes into `_rfc5_transform` when possible (e.g. with `hcs.py`) and we reimplement them from scratch when not possible (e.g. with `multiscales.py`). We did this to minimize code duplication.
2. in `spatialdata`, [we implemented](https://github.com/scverse/spatialdata/blob/main/src/spatialdata/transformations/ngff/ngff_transformations.py) almost all the NGFF transformation classes. I didn't move code from there yet because first I wanted an initial example to agree on the code structure. That would be one of the next steps.

## Conclusions
Thanks for the review and hit us up for any question!


src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+        Number of dimensions.
+        """
+        return len(self.scale)
+
+
+class Translation(CoordinateTransformation):
+    """Translation transformation."""
+
+    type: Literal["translation"] = "translation"
+    translation: list[float]
```
d-v-b, line 74: `Translation.translation` hurts my eyes, but I guess that's what people wanted.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+        Number of dimensions.
+        """
+        return len(self.translation)
+
+
+class Sequence(CoordinateTransformation):
+    """Sequence transformation."""
+
+    type: Literal["sequence"] = "sequence"
+    transformations: tuple[CoordinateTransformationType, ...]
```
d-v-b, line 88: why is this `tuple` when `Translation.translation`  and `Scale.scale` are lists?

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+CoordinateTransformationType = Annotated[
```
d-v-b, line 91: is the discriminator necessary? I think all of these classes have separable fields, so they should be structurally disambiguous without a special annotation.

src/ome_zarr_models/_rfc5_transforms/image.py
```diff
@@ -0,0 +1,24 @@
...
+
+__all__ = ["Image", "ImageAttrs"]
+
+
+class ImageAttrs(BaseOMEAttrs):
+    """
+    Model for the metadata of OME-Zarr data.
+    """
+
+    multiscales: list[Multiscale] = Field(
```
d-v-b, line 14: there's an RFC that makes multiscales singular. Not sure if that goes before or after the transformations though.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+    ) -> object:
+        """
+        Ensures that
+        - a single transformation is present
+        - such transformation is a scale
+        - if such transformation is a sequence, ensure that its length is 2 and that
+          the first transformation is a scale and the second a translation
+        """
+
+        class Transforms(BaseModel):
```
d-v-b, line 51: why are we defining a class inside a validator function?

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+            transforms: tuple[CoordinateTransformationType, ...]
+
+        transforms = Transforms(transforms=transforms_obj).transforms
```
d-v-b, line 54: this is too many "transforms" 😆 are you just using this class for validation? because that's why it's better IMO to define validator functions separately from the classes, so you can re-use them.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+    # TODO: suggest John not to have a tuple for the coordinate transformations of a
+    #  dataset, since it's either a single scale or a single sequence (with a scale
+    #  and a translation)
+    coordinateTransformations: tuple[CoordinateTransformationType, ...] = Field(
+        ..., min_length=1, max_length=1
+    )
+
+    # the before validation is used to simplify the error messages
+    @field_validator("coordinateTransformations", mode="before")
+    def _ensure_scale_translation(
```
d-v-b, line 40: this method probably needs to be renamed

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+            if (
+                isinstance(first, Scale)
+                and isinstance(second, Translation)
+                and len(first.scale) != len(second.translation)
+            ):
+                raise ValueError(
+                    "The length of the scale and translation vectors must be the same."
+                    f"Got {len(first.scale)} and {len(second.translation)}."
+                )
+        return transforms
```
d-v-b, line 112: is this dimensionality validation not already part of the sequence class definition? i.e., once we know it's a sequence, haven't we already checked that the transforms inside the sequence are dimensionally compatible?

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: I got used to this, but I see what you mean.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: Good point, we may have made this mistake elsewhere.

src/ome_zarr_models/_rfc5_transforms/image.py
LucaMarconato: True, maybe easier to implement that RFC before.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+    @field_validator("coordinateTransformations", mode="before")
+    def _ensure_scale_translation(
+        transforms_obj: object,
+    ) -> object:
+        """
+        Ensures that
+        - a single transformation is present
+        - such transformation is a scale
+        - if such transformation is a sequence, ensure that its length is 2 and that
+          the first transformation is a scale and the second a translation
```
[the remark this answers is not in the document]
LucaMarconato: I also saw this as a bit ugly.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: Good point. I implemented this because `Sequence` is still empty, but it's better to move the validation there.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+                )
+        return transforms
+
+
+class Multiscale(BaseAttrs):
+    """
+    An element of multiscales metadata.
+    """
+
+    coordinateSystems: tuple[CoordinateSystem, ...] = Field(..., min_length=1)
```
d-v-b, line 120: does the spec allow multiple coordinate systems with the same `name` attribute?

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+
+from pydantic import Field, field_validator
+
+from ome_zarr_models._rfc5_transforms.axes import Axes
+from ome_zarr_models._utils import duplicates
+from ome_zarr_models.base import BaseAttrs
+
+
+class CoordinateSystem(BaseAttrs):
+    name: str
```
dstansby, line 13:
```suggestion
    name: str = Field(min_legnth=1)
```
This is much simpler than the custom validator below

dstansby:
Thanks a bunch for this! I'm going to start reviewing by commenting on the code structure, and I'll comment as I go so since it'll take me a while to go through everything. Once we've got the structure sorted, I'll turn to checking the tests and the implementation against the draft RFC5.

Thanks again for this - it's a ton of work, and I really appreciate it!

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @field_validator("axes")
+    def axes_must_not_be_empty(cls, value: Axes) -> Axes:
+        if len(value) == 0:
+            raise ValueError("axes must contain at least one axis")
+        return value
+
+    @field_validator("axes", mode="after")
+    @classmethod
+    def _ensure_unique_axis_names(cls, axes: Axes) -> Axes:
```
dstansby, line 30: There's already `ome_zarr_models.common.validation.unique_items_validator` that I think would work here.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
dstansby:
```suggestion
    transformations: tuple[CoordinateTransformation, ...]
```
Is it not possible just to use the base class here, instead of defining a new type?

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+from ome_zarr_models._utils import duplicates
+from ome_zarr_models.base import BaseAttrs
+
+
+class CoordinateSystem(BaseAttrs):
+    name: str
+    axes: Axes
+
+    @field_validator("name")
+    def name_must_not_be_empty(cls, value: str) -> str:
```
dstansby, line 17: Can you just use a `Field(..., min_length=1)` above, instead of defining this custom validator?

src/ome_zarr_models/_rfc5_transforms/multiscales.py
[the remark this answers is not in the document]
dstansby: Can we suggest on the RFC that it's always (scale, translation), and if you just want a translation you should set the scale values explicitly to 1?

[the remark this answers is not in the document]
d-v-b: i think the fundamental problem is with the ambiguity between a sequence of transforms with 1 transform, vs a single transform outside of a sequence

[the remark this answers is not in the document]
will-moore: It feels like this is hard because we're using a `coordinateTransformations: []` list on the `dataset`. If we just want to be sure that the dataset has a scale transformation, maybe the dataset should just have a single `scaleCoordinateTransformation: {}` and an optional `translationCoordinateTransformation: {}` and that's it.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
d-v-b: i checked, and the rfc5 states that the `name` attributes must be unique across elements of `coordinateSystems`. this means that a tuple is actually unnecessary, and this should be a dict instead. but that requires changing rfc5.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+class Dataset(BaseAttrs):
+    """
+    An element of Multiscale.datasets.
+    """
+
+    path: str
+    # TODO: suggest John not to have a tuple for the coordinate transformations of a
+    #  dataset, since it's either a single scale or a single sequence (with a scale
+    #  and a translation)
+    coordinateTransformations: tuple[CoordinateTransformationType, ...] = Field(
```
d-v-b, line 34: according to the current text of rfc5, only scale and translation are allowed in here. so a sequence that contains a scale / translation is not permitted. (I don't really understand why this requirement exists)

d-v-b: see https://github.com/bogovicj/ngff/blob/52d7924f1522bdf917ea912fc416ae55b3229ebb/0.6-dev/index.bs#L1366-L1368

src/ome_zarr_models/_rfc5_transforms/image.py
dstansby: Yeah, it's https://github.com/ome/ngff/pull/285, but I don't have much hope for it going anywhere soon... so I think lets leave this as is. Easy enough to change it later.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
dstansby: My comment on the existing code is

```python
        # This is used as a before validator - to help use, we use pydantic to first
        # cast the input (which can in general anything) into a set of transformations.
        # Then we check the transformations are valid.
        #
        # This is a bit convoluted, but we do it because the default pydantic error
        # messages are a mess otherwise
```

I'd suggest copy/pasting/modifying that here, to avoid confusion because defining the class here is a bit weird.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+    type: JsonValue = None
+
+    @property
+    def ndim(self) -> int:
+        """
+        Dimensionality of the data described by this metadata.
+
+        Determined by the length of the axes attribute.
+        """
+        output_cs_name = self.datasets[0].coordinateTransformations[0].output
```
dstansby, line 134: Instead of taking the first dataset here, shouldn't `ndim` be a property on `Dataset` instead (because different datasets in multiscales could have different dimensionalities)?

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+    def _ensure_same_output_cs_for_all_datasets(data: Self) -> Self:
+        """
+        Ensure that all datasets have the same output coordinate system.
+
+        Also ensures that the dimensionality of each dataaset match the one of the
+        output coordinate system.
+        """
+        output_cs_names = set()
+        for dataset in data.datasets:
+            output_cs_names.add(dataset.coordinateTransformations[0].output)
```
dstansby, line 153:
```suggestion
        output_cs_names = {dataset.coordinateTransformations[0].output for dataset in datasets}
```

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,247 @@
...
+        if len(output_cs_names) > 1:
+            raise ValueError(
+                "All `Dataset` instances of a `Multiscale`  must have the same output "
+                "coordinate system. Got {output_cs_names}."
```
dstansby, line 157:
```suggestion
                f"coordinate system. Got {output_cs_names}."
```

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+        if len(name_dupes) > 0:
+            msg = (
+                f"Axis names must be unique. Axis names {tuple(name_dupes.keys())} are "
+                "repeated."
+            )
+            raise ValueError(msg)
+        return axes
+
+
+class CoordinateTransformation(BaseAttrs):
```
d-v-b, line 44: A few complications for `input` and `output`, based on a close read of the current version of the transforms PR:
- in the context of a `byDimension` transform, `input` / `output` could be a _list_ of strings. This may vastly complicate validation. See discussion [here](https://github.com/ome/ngff/pull/138/files#r2081982273).
- `input`/`output` are either a string / list of strings, or unset. So this type should include some logic that excludes those fields from JSON serialization if they are set to `None`.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+class Identity(CoordinateTransformation):
+    """Identity transformation."""
+
+    type: Literal["identity"] = "identity"
+
+
+class Scale(CoordinateTransformation):
+    """Scale transformation."""
+
+    type: Literal["scale"] = "scale"
```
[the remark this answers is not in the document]
d-v-b: since the "scale with path" and "scale with scale" transforms are disjoint, I originally implemented them as two separate classes, which can simplify some of the validation

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def ndim(self) -> int:
+        """
+        Number of dimensions.
+        """
+        return len(self.scale)
+
+
+class Translation(CoordinateTransformation):
```
d-v-b, line 70: similar to the [`Scale` transform](https://github.com/ome-zarr-models/ome-zarr-models-py/pull/187/files#r2083059679), this is missing support for the `path` field.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
d-v-b: should the `Sequence` type apply validation on the elements of `transformations`? I think the spec states that the transformations must have compatible dimensionality, for example.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: I'll resolve this, but if the specs is tweaked we can always easily change this.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: > should the Sequence type apply validation on the elements of transformations? I think the spec states that the transformations must have compatible dimensionality, for example.

Yes, this was on purpose omitted from this first version. I will open an issue so we don't forget about this.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
d-v-b: oh also, according to Each dictionary in "datasets" MUST contain the field "coordinateTransformation", whose value is a dictionary that defines a, this field is called `coordinateTransformation`, not `coordinateTransformations`. `coordinateTransformations` is stored in zarr group attrs I think? These names are super confusing!

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: > Is it not possible just to use the base class here, instead of defining a new type?

@dstansby I tried replacing `CoordinateTransformationType` with `CoordinateTransformation` here (and in all the places where it is imported but one needs to make some extra changes. Here what I want to achieve is:
1) inheriting validation logic from `CoordinateTransformation`, like for `input` and `output`. 
2) forbid the user to have an object of type `CoordinateTransformation`. 

Usually I'd use an abstract class with a factory class method. Here I tried to achieve this with a discriminated union type instead of the factory method. I tried quickly using a factory method but the code was getting more complex than the one I implemented. Can you try please?

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: Good point. I replaced it with `CoordinateTransformationType = Identity | Scale | Translation | Sequence`

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: I agree, quite a weird line and having a helper function would be better. The problem is that the two `class Transform` are using different `CoordinateTransformation` classes (one from `common` and the other one from `_rfc5_transforms`). So in practice, maybe the current is already one of the simplest/shortest codes we can have.

I'd leave it as it is, but if you want to try a simplification please you are welcome 😊

src/ome_zarr_models/_rfc5_transforms/multiscales.py
d-v-b: another issue with the spec: it states that the name of an array can be used to denote that array's coordinate system. but it's unclear what should happen if the name of the array is already used by an element of `coordinateSystems`.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: Any suggestion? A more explicit name could be `_ensure_scale_or_sequence_of_scale_translation()`. But maybe the current one is good enough (and consistent with `common.multiscale.Dataset`).

d-v-b: i don't think scale + translation is [required any more](https://github.com/bogovicj/ngff/blob/3822dd0d9d2e388a12b6b74d0bffc7a215298e42/0.6-dev/index.bs#L1348-L1351), so maybe you can get rid of this method entirely?

LucaMarconato: But so here now we could have an affine or other things? If this is the case, then a `Multiscale` object would go beyond a pyramidal image. It could represent a tiled image, or a set of 2 images in a 3D space, etc.

Is this what the new specs want?

src/ome_zarr_models/_rfc5_transforms/multiscales.py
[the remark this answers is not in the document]
LucaMarconato: Resolving this in favor of this comment here: https://github.com/ome-zarr-models/ome-zarr-models-py/pull/187#discussion_r2087541247

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: I agree with both and thanks for pointing this out in Zulip. Could you please open a GitHub issue for both? I am afraid that otherwise these important points may get lost.

LucaMarconato: Actually, I created such issue: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/189.

LucaMarconato: I would resolve this conversation since the code todo items can be discussed in the linked issue and tackled in a follow up PR.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: Thanks! Simplified.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: 👍

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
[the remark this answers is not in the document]
LucaMarconato: Thanks! I omitted this on purpose from this PR to attack a smaller problem. I opened an issue to track this but I would do this in a follow up PR: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/190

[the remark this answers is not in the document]
LucaMarconato: I'll resolve the conversation.

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: See comments here: https://github.com/ome-zarr-models/ome-zarr-models-py/pull/187#discussion_r2083059679. Resolving the conversation.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: I will change the name from `coordinateTransformations` to `coordinateTransformation` for distinguishing the transformation in `Dataset` from the transformation in `Multiscale`.

LucaMarconato: Actually I checked the examples and the name is actually `coordinateTransformations` and not `coordinateTransformations`. I commented this here https://github.com/ome/ngff/pull/138#discussion_r2087612094.

I would not change the code unless the official examples change. But if turns out not to be a typo, then we can update both the examples and code.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: In the old version of transformations all datasets had the same dimensionality (this was validated here https://github.com/ome-zarr-models/ome-zarr-models-py/blob/0f7fb161b4bd8d34e1b7106cbb5fcfc6c50910c0/src/ome_zarr_models/common/multiscales.py#L169)

Now in the new specs (see also https://github.com/ome-zarr-models/ome-zarr-models-py/pull/187#discussion_r2087541247), has this changed? As in the linked comment, is this what the new specs want?

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
LucaMarconato: I agree, the `byDimension` case complicates the handling of `input` and `output`. I think we should deal with this in a separate PR. I moved this discussion in a issue: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/191.
I will resolve this conversation.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
dstansby: I'm re-reading the spec as altered by RFC 5, and I think it implies (but doesn't explicitly say) that all datasets have the same dimensionality, so we can assume that. So I think the code as written is valid/fine.

It might be a bit simpler to get `ndim` from `self.coordinateSystems[0]` here though?

dstansby (approved):
Looking good! I left some final comments/suggestions. I think a way forward would be to address those, then we can merge this. Do you plan on following this up with more PRs? Once everything is done I should probably go through and double check it all against the new spec (as modified by RFC 5), but I will do that after you've finished with all your (awesome!) work.

dstansby:
👋 @LucaMarconato now OME-Zarr 0.5 support is released, I'd like to work on implementing RFCs. I tried re-reading this PR and it's quite a big one to go over, so I was wondering how you would feel about implementing it in smaller chunks, corresponding to the sections in https://ngff.openmicroscopy.org/rfc/5/index.html#proposal. For example, a first PR could implement the "“coordinateSystems” metadata" section and related tests, pulling out that functionality from this PR. It is a bit more work this way, but will be much easier for me or someone else to review. Do let me know if you'll have any time to work on this in the next few weeks or not too - if not, I can try take your excellent work here and split it up into smaller PRs to merge.

dstansby:
Okay, that sounds good 👍 I forgot that a lot of this PR is tests instead of implementation 🤦 , which should make it a lot easier to review. So if you go ahead and let me know when this is ready for a quick review, I will give it another look 😄 

Note that I created a v06 implementation (https://github.com/ome-zarr-models/ome-zarr-models-py/pull/261) that this work should eventually be targeted against. We can move the modifications out of your `_rfc_5_transformations` folder and against the `v06` folder later though. Sorry for the confusion and different folders here. It's been a bit of a trial and error working out the best way to do this, as it's the first time an RFC has been implemented. But I think we've settled on a good way to do it now.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
LucaMarconato: Opened an issue for this here: https://github.com/ome/ngff/issues/331

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,87 @@
...
+    @classmethod
+    def _ensure_unique_axis_names(cls, axes: Axes) -> Axes:
+        """
+        Ensures that the names of the axes are unique.
+        """
+        unique_items_validator([a.name for a in axes])
+        return axes
+
+
+class CoordinateTransformation(BaseAttrs, ABC):
```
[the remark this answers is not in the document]
LucaMarconato: Added!

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,87 @@
...
+        return self
+
+
+class Identity(CoordinateTransformation):
+    """Identity transformation."""
+
+    type: Literal["identity"] = "identity"
+
+
+class Scale(CoordinateTransformation):
```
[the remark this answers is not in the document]
LucaMarconato: Exactly, tracked here: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/190

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Sequence(CoordinateTransformation):
+    """Sequence transformation."""
+
+    type: Literal["sequence"] = "sequence"
+    transformations: list[CoordinateTransformationType]
+
+
+CoordinateTransformationType = Identity | Scale | Translation | Sequence
```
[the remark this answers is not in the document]
LucaMarconato: Yes, here we start with basic ones (enough to represent multiscale objects).

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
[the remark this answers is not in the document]
LucaMarconato: Good point!

I think you meant "I would suggest renaming this to CoordinateTransformationType," instead of "I would suggest renaming this to CoordinateTransformation,".

I will do the following renamings:
- `CoordinateTransformation(BaseAttrs)` -> `CoordinateTransformationBase(BaseAttrs)`
- `CoordinateTransformationType` -> `CoordinateTransformation`.

src/ome_zarr_models/_rfc5_transforms/multiscales.py
```diff
@@ -0,0 +1,251 @@
...
+                )
+        return transforms
+
+
+class Multiscale(BaseAttrs):
+    """
+    An element of multiscales metadata.
+    """
+
+    coordinateSystems: tuple[CoordinateSystem, ...] = Field(..., min_length=1)
```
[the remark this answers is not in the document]
LucaMarconato: Added!

src/ome_zarr_models/_rfc5_transforms/coordinate_transformations.py
```diff
@@ -0,0 +1,87 @@
...
+from pydantic import Field, field_validator, model_validator
+
+from ome_zarr_models._rfc5_transforms.axes import Axes
+from ome_zarr_models.base import BaseAttrs
+from ome_zarr_models.common.validation import unique_items_validator
+
+
+class CoordinateSystem(BaseAttrs):
+    name: str = Field(min_length=1)
+    axes: Axes = Field(min_length=1)
```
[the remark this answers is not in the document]
LucaMarconato: Yes, tracked here: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/265

LucaMarconato:
@will-moore this PR uses the data you mentioned as a reference and also the examples from the John's branch of the NGFF repo. Please see more here on how this PR gets and uses the data: https://github.com/melonora/ome-zarr-models-py/tree/rfc5/tests/_rfc5_transforms/data_rfc5.

Please note that the purpose of this PR is to setup tests against such data and to add initial validation tests only for simple transformations (multiscale images and sequences of scale + translation). 

If we find errors in the data (as the one you found), we will make a PR to the data repos above to propose a fix (please also can you make a PR for the problems you discovered).

will-moore:
@LucaMarconato Thanks - I was reading https://github.com/melonora/ome-zarr-models-py/tree/rfc5/tests/_rfc5_transforms/data_rfc5 earlier but now it's 404!
Ah - found it! https://github.com/melonora/ome-zarr-models-py/tree/rfc5/tests/data/examples/v06#readme

I followed the various steps which seemed to work until the end  - ls gave me everything!

```
$ ls
_static			0.5			CONTRIBUTING.md		latest			README.md
0.1			0.6-dev			data			LICENSE.md		requirements.txt
0.2			about			environment.yaml	make.bat		rfc
0.3			community		images			Makefile		specifications
0.4			conf.py			index.rst		publications		tools
```

But how do I actually run the tests? I tried just `pytest` but no joy

```
$ pwd
/Users/wmoore/Desktop/OME-ZARR-MODELS/ome-zarr-models-py/tests
$ pytest _v06
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov --cov-report=xml
  inifile: /Users/wmoore/Desktop/OME-ZARR-MODELS/ome-zarr-models-py/pyproject.toml
  rootdir: /Users/wmoore/Desktop/OME-ZARR-MODELS/ome-zarr-models-py
```

will-moore:
Thanks, that worked great. 

```
$ pytest _v06
...
== 3 failed, 57 passed, 64 skipped, 4 xfailed in 2.66s ===
```

LucaMarconato:
I actually simplified a bit the tests handling and removed the need to use `sparseCheckout`, by adding `--ignore=tests/data/examples/v06/ngff/tests` to `pyproject.toml`. Could you try again please?

Now it should be as easy as cloning with the submodules, and all tests should pass.

will-moore:
```
$ git submodule update
Submodule path 'tests/data/examples/v06/ngff': checked out '07ecd1a29603544a7073bc8cbe8f74e70ec8408a'
$ cd tests
$ pytest _v06
...
=== 59 passed, 64 skipped, 4 xfailed in 1.79s ===
```
👍

LucaMarconato:
> Note that I created a v06 implementation (https://github.com/ome-zarr-models/ome-zarr-models-py/pull/261) that this work should eventually be targeted against. We can move the modifications out of your _rfc_5_transformations folder and against the v06 folder later though.

@dstansby this is now addressed and all the tests pass. 

Below is a overview of this PR. I think now that we build upon `v06` it should be much easier to re-review and probably the PR would not need to be split. In such case [we still need to decide how to proceed for these 2 open issues](https://github.com/ome-zarr-models/ome-zarr-models-py/pull/187#issuecomment-3248777794), but then if you agree, after the re-review we could merge and continue the work in smaller follow up PRs. 

# List of files that changed in this PR with significant changes.
```
.
├── src
│   └── ome_zarr_models
│       ├── _v06
│       │   ├── coordinate_transformations.py.  <--- new!
│       │   ├── image.py  <--- almost untouched + commented out code left for later
│       │   └── multiscales.py  <--- new!
├── tests
│   ├── _v06
│   │   └── conftest.py  <--- new long file with some complexity, commented below
│   └── test_data_rfc5
│       ├── ngff  <--- just tests; it's enough to look at "test_identity()"
│       ├── ngff_rfc5_coordinate_transformation_examples  <--- just tests, and all skipped at the moment
│       ├── test_all_data_is_tested.py <--- the external data will change, and this test will make sure that are tests are complete
│       ├── test_image.py <--- the v05 code was easy to port to v06, I commented some old functions (from v05) that needs to be adapted
│       ├── test_image_label.py <--- small fixes
│       ├── test_multiscales.py <--- new tests! + old tests that needs from v05 to be adapted (for the moment commented out)
│       └── test_transformations.py <--- new!
└── data
    └── examples
        └── v06
            ├── image_example.json <--- small fixes
            ├── image_label_example.json <--- as above
            ├── image_no_version_example.json <--- as above
            ├── labels_image_example.json <--- as above
            ├── ngff <--- submodule
            ├── ngff-rfc5-coordinate-transformation-examples <--- submodule
            └── spec_example_multiscales.json <--- I removed this because it looked unused
```

LucaMarconato:
As shown in the summary above, only 3 .py files, and 3 .py tests files, are entirely new. 

Among these, here are some comments on `tests/_v06/conftest.py` and `tests/_v06/test_data_rfc5/test_all_data_is_tested.py`.
These two files have some verbosity and complexity, due to two reasons:
1. The examples in `tests/data/examples/v06/ngff/examples` are not full examples as the new ones from the `ngff-rfc5-coordinate-transformation-examples` repo, so to test them I need to embed them into a full object. Most of the complexity in `conftest.py` is doing this.
2. The rests of the code in `conftest.py` and the code in `test_all_data_is_tested.py` is ensuring that as the external data will change (and we expect to have it changing often), we are still testing all the functions and not missing anything.

will-moore:
@LucaMarconato I opened a PR for the easy fixes at https://github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples/pull/12

And opened an issue for the other point (which I don't know how to fix) https://github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples/issues/13

tests/data/examples/v06/image_example.json
```diff
@@ -4,20 +4,38 @@
...
-        "axes": [
-          { "name": "t", "type": "time", "unit": "millisecond" },
-          { "name": "c", "type": "channel" },
-          { "name": "z", "type": "space", "unit": "micrometer" },
-          { "name": "y", "type": "space", "unit": "micrometer" },
-          { "name": "x", "type": "space", "unit": "micrometer" }
+        "type": "gaussian",
+        "coordinateSystems": [
+          {
+            "name": "example",
```
dstansby, line 10: Could we rename this "coord_sys_1" or similar so it's a bit less generic than just "example" (and the same for other coordinate systems in the metadata files)

tests/data/examples/v06/image_example.json
```diff
@@ -4,20 +4,38 @@
...
         ],
         "datasets": [
           {
             "path": "0",
             "coordinateTransformations": [
               {
                 "type": "scale",
-                "scale": [1.0, 1.0, 0.5, 0.5, 0.5]
+                "scale": [1.0, 1.0, 0.5, 0.5, 0.5],
+                "input": "/0",
```
dstansby, line 37: Shouldn't this be a coordinate system name, not a path to an image?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -1,9 +1,261 @@
...
+    metadata: JsonValue = None
+    name: JsonValue | None = None
+    type: JsonValue = None
+
+    @property
+    def ndim(self) -> int:
+        """
+        Dimensionality of the data described by this metadata.
+
+        Determined by the length of the axes attribute.
```
dstansby, line 136:
```suggestion
```
This is no longer true :)

dstansby (approved):
I've given this a review and it's looking really nice! My main worry is about how complex the test machinery you set up is. I think it's justified, but it would be nice if there were some more docstrings explaining what's going on (I left some comments to this effect inline).

Once you've had a look and possibly addressed my comments, feel free to merge - I invited you to https://github.com/orgs/ome-zarr-models/teams/ome-zarr-models-developers/members, and if you accept the invitation you should get merge rights to this repository.

Thanks again for all this work - I'm excited to keep the momentum going now!

d-v-b:
instead of the git submodules, maybe it would be easier to just fetch stuff from github as regular files?

src/ome_zarr_models/_v06/coordinate_transformations.py
```diff
@@ -1,21 +1,106 @@
...
+        Ensures that the names of the axes are unique.
+        """
+        unique_items_validator([a.name for a in axes])
+        return axes
+
+
+class CoordinateTransformationBase(BaseAttrs):
+    type: str
+    input: str | None = None
+    output: str | None = None
```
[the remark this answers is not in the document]
LucaMarconato: Yes, we need to add this validation. I actually have code for this in `spatialdata`. I will add a GitHub issue to not forget but I'd move the code later.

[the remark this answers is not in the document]
LucaMarconato: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/266

tests/data/examples/v06/image_example.json
LucaMarconato: Good point.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -1,9 +1,261 @@
...
+class Dataset(BaseAttrs):
+    """
+    An element of Multiscale.datasets.
+    """
+
+    path: str
+    # TODO: suggest John not to have a tuple for the coordinate transformations of a
+    #  dataset, since it's either a single scale or a single sequence (with a scale
+    #  and a translation)
+    coordinateTransformations: list[CoordinateTransformation] = Field(
```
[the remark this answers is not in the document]
LucaMarconato: I will leave it like this for the moment for the following reasons:
- I'd need to change all the hardcoded examples from list to tuple (and v0.4 and v0.5 uses lists)
- I'd still need to have validation for the elements of `SequenceTransform`, so it would be less code, but still some code.
- It's verbose in terms of code, but the error messages are nice.

We can wait to see if John removes the need to have a single-element list or a single-element tuple and then simplify.

src/ome_zarr_models/_v06/multiscales.py
LucaMarconato: Good point!

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -1,9 +1,261 @@
...
+        Dimensionality of the data described by this metadata.
+
+        Determined by the length of the axes attribute.
+        """
+        output_cs_name = self.datasets[0].coordinateTransformations[0].output
+        output_cs: CoordinateSystem | None = None
+        for cs in self.coordinateSystems:
+            if cs.name == output_cs_name:
+                output_cs = cs
+                break
```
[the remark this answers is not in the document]
LucaMarconato: Done.

LucaMarconato:
Thanks @dstansby for the review! I am now simplifying the testing machinery. Most of the complexity comes from testing against [these json files](https://github.com/bogovicj/ngff/tree/dev/examples/transformations), not being complete examples. Maybe we should move them into complete examples in the `ngff-rfc5-coordinate-transformation-examples` and get rid of complexity. But we can do this any time, so I'll leave things like this for the moment.

LucaMarconato:
> instead of the git submodules, maybe it would be easier to just fetch stuff from github as regular files?

My first implementation was downloading the files. I'd see if we are comfortable with the current submodule system, and if not we can replace that with downloading files. In both cases we avoid the need to having to commit the files to the repo, and in both cases syncing against the latest changes is easy.

One drawback of submodules is that the GitHub workflow to test the data may need some extra lines of code, I can check this.

dstansby:
Since it's going to be mainly the three of us discussing here working on RFC5, as long as we're happy with the setup I don't think the complications of submodules matters to much. When RFC5 (finally!) settles down into a "responded to reviewers" state, and we merge the `rfc5` branch into `main`, we can copy/paste across the test data and get rid of the submodules to make it easier for other contributors.

will-moore:
Any plans to have the tests run as part of the github actions on every push? I don't know if thats any harder to setup if you're using git submodules but probably not much

dstansby:
Huh, the tests *should* be running! I suspect this is because this PR is targeting a non-`main` branch, and the GH actions config needs updating to take that into account.

PR #195, opened 2025-05-19
Add .datasets property to Image

dstansby:
I'm not 100% sure this is a good idea, so feedback encouraged and requested.

To get the (e.g.,) first level of a multiscales, currently the syntax is
```python
ome_group[ome_zarr_image.attributes.multiscales[0].datasets[0].path]
```

this helper would shorten this to

```python
ome_group[ome_zarr_image.datasets[0][0].path]
```

### DIFF
--- a/src/ome_zarr_models/v04/image.py
+++ b/src/ome_zarr_models/v04/image.py
@@ -221,15 +221,15 @@ class Image(BaseGroupv04[ImageAttrs]):
 
         return Labels(attributes=labels_group.attributes, members=labels_group.members)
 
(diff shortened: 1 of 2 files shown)


LucaMarconato:
Why the double `[0]` in `ome_group[ome_zarr_image.datasets[0][0].path]`?

dstansby:
The first index is the index of the multiscales (because multiscales is an array and not just a single object (at the moment... https://ngff.openmicroscopy.org/rfc/6/index.html))

LucaMarconato:
Thanks! But then shouldn't there be an extra `0` somewhere in `ome_group[ome_zarr_image.get_datasets()[0].path]`?

PR #197, opened 2025-05-25
Remove zarr dependency

dstansby:
This is an experiment to see if it's possible to remove our zarr dependency. I think it might be possible if we:

- Move some of the `import zarr`s into type checking blocks
- import zarr inside functions that use it

We would need to add a new GitHub actions job that runs the tests without zarr installed to check this is working.


dstansby:
This is going to require the same change to `pydantic-zarr` first, so I think I'lll close this in the meantime.

PR #203, opened 2025-06-06
Change default coordinateTransforms to an array

dstansby:
The 0.4 spec says:

> Each "multiscales" dictionary MAY contain the field "coordinateTransformations", describing transformations that are applied to all resolution levels in the same manner. The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations" and are applied after them. 

> It (coordinateTransformations) is a list of dictionaries.

So I think by having `null` as a default is actually not compliant, but having an empty array should be compliant.

At least with `neuroglancer`, having `null` errors, but having an empty array is fine.


d-v-b:
it would be good to test this behavior with other viewers. since the spec is so imprecise, it wouldn't surprise me if there's a fair amount of skew across implementations here

dstansby:
> None is a fine default provided we don't serialize it to JSON. In this case, None means "this key will be missing".

I don't think (I did look) there's an elegant way to mark a field as "only serialise if not None".

I think this is a pragmatic solution to fix data to work with `neuroglancer` for now, and I also think it's nice to be explicit about including an empty array here, to communicate "this dataset has no global transformations" if you're reading the OME-Zarr serialised attributes.

d-v-b:
> > None is a fine default provided we don't serialize it to JSON. In this case, None means "this key will be missing".
> 
> I don't think (I did look) there's an elegant way to mark a field as "only serialise if not None".

I solved this problem over in pydantic_ome_ngff. First, you list all the fields that should not be serialized if they are `None` with a dedicated class attribute. Then, when serializing the model, you skip any keys that are listed in that class attribute iff they are `None`. Here's the [core machinery](https://github.com/JaneliaSciComp/pydantic-ome-ngff/blob/06dc31cfb734d278cd52027d1b9e633edc0e67b7/src/pydantic_ome_ngff/base.py#L23-L64), and an [example of its usage](https://github.com/JaneliaSciComp/pydantic-ome-ngff/blob/06dc31cfb734d278cd52027d1b9e633edc0e67b7/src/pydantic_ome_ngff/v04/multiscale.py#L214-L218).

d-v-b:
> I also think it's nice to be explicit about including an empty array here, to communicate "this dataset has no global transformations" if you're reading the OME-Zarr serialised attributes.

I agree that it's good to explicitly signal the lack of global transformations, but the spec defines an implicit way (via the missing key) and an explicit way (via the empty array). this library should support both.

d-v-b:
also, neuroglancer is wrong to accept the empty array here. the spec says:

`The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations" and are applied after them.` The datasets definition of coordinateTransformations requires either a `[scale]` or `[scale, translation]` array. An empty array is not allowed.

dstansby:
> I agree that it's good to explicitly signal the lack of global transformations, but the spec defines an implicit way (via the missing key) and an explicit way (via the empty array). this library should support both.

👍  - do you think we shouldn't make the change here without supporting the missing key way, or do you think we should also support the missing key at the same time?

My thoughts are this is stops invalid data (`null`) being written, it's a simple fix, and it doesn't preclude supporting the missing key later, so I'm minded to merge this as a quick bugfix and deal with the missing key stuff across all data structures consistently later.

PR #209, opened 2025-07-22
Convert v05 to use Zarr-python 3

dstansby:
This gets OME-Zarr v0.5 working with Zarr-python 3.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,6 @@ docs = [
 [tool.hatch.version]
 source = "vcs"
 
-[tool.hatch.metadata]
-allow-direct-references = true
-
 [build-system]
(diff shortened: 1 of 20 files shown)


dstansby:
This is almost working 🎉  - @d-v-b is there any chance you could take a look at the `mypy` errors? I could stick an ignores on them, but thought it was worth a second pair of eyes seeing if they're solveable.

pyproject.toml
```diff
@@ -35,6 +35,9 @@ docs = [
 [tool.hatch.version]
 source = "vcs"
 
+[tool.hatch.metadata]
+allow-direct-references = true
```
d-v-b, line 39: do we still need this?

dstansby: Woops, good catch

PR #210, opened 2025-07-23
Replace create_dataset with create_array

dstansby:


dstansby:
Never mind, this won't work with zarr-python 2

Issue #211, opened 2025-07-24
plate missing subgroups inconsistent


will-moore:
Just testing metadata, as described at https://github.com/ome/ome-zarr-py/issues/460.

Strange that the missing `A` subgroup only causes a validation error when we have `acquisitions` in the plate metadata.

```
ome-zarr-models==0.1.8
pydantic-zarr==0.7.0
zarr==2.18.4
```

```
import zarr
from ome_zarr_models.v04 import HCS as Models04HCS

plate_group = zarr.open("test_plate_acquisition.zarr", mode="w")

plate_group.attrs["plate"] = {
    "columns": [{"name": "1"}],
    "field_count": 10,
    "rows": [{"name": "A"}],
    "version": "0.4",
    "wells": [{"columnIndex": 0, "path": "A/1", "rowIndex": 0}]
}
# No error here... even though subgroup "A" does not exist
Models04HCS.from_zarr(plate_group)

plate_group.attrs["plate"] = {
    "acquisitions": [{"id": 1}, {"id": 2}, {"id": 3}],
    "columns": [{"name": "1"}],
    "field_count": 10,
    "rows": [{"name": "A"}],
    "version": "0.4",
    "wells": [{"columnIndex": 0, "path": "A/1", "rowIndex": 0}]
}
# Gives KeyError: 'A'
Models04HCS.from_zarr(plate_group)
```

will-moore:
It doesn't depend on external data. You can run the test as it is. It creates the data.

PR #214, opened 2025-07-30
Allow HCS objects without existing well groups

dstansby:
The relevant part of the 0.4 specificaiton says:

> The plate dictionary MUST contain a wells key whose value MUST be a list of JSON objects defining the wells of the plate. Each well object MUST contain a path key whose value MUST be a string specifying the path to the well subgroup. The path MUST consist of a name in the rows list, a file separator (/), and a name from the columns list, in that order. The path MUST NOT contain additional leading or trailing directories. Each well object MUST contain both a rowIndex key whose value MUST be an integer identifying the index into the rows list and a columnIndex key whose value MUST be an integer identifying the index into the columns list. rowIndex and columnIndex MUST be 0-based. The rowIndex, columnIndex, and path MUST all refer to the same row/column pair.

There's no stipulation that the path exists or points to a valid Well Zarr group, so don't error out if some well groups don't exist.

cc @will-moore - can you take a look at this and let me know if it fixes your issue?

Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/211.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.1.10
+
+### Bug fixes
+
+- Creating a `HCS` group with well paths that do not point to existing well Zarr groups
+  no longer errors.
+
 ## 0.1.9
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


will-moore (approved):
Thanks, that looks good and fixes the error I was seeing at #211

Issue #236, opened 2025-08-20
`Image.labels` property fails


lorenzocerrone:
Accessing the `Image.labels` property fails for images with labels.

Minimal example:

```python
import json

from ome_zarr_models.v05.image import Image
from tests.v05.conftest import json_to_zarr_group

image_group = json_to_zarr_group(
    json_fname="/Users/locerr/Projects/ome-zarr-models-py/tests/v05/data/image_example.json"
)
image_group.create_array(
    "0",
    shape=(1, 1, 1, 1, 1),
    dtype="uint8",
)
image_group.create_array("1", shape=(1, 1, 1, 1, 1), dtype="uint8")
image_group.create_array("2", shape=(1, 1, 1, 1, 1), dtype="uint8")

labels_group = image_group.create_group(name="labels")
labels_group.attrs.put({"ome": {"labels": ["test"], "version": "0.5"}})

test_group = labels_group.create_group("test")

with open(
    "/Users/locerr/Projects/ome-zarr-models-py/tests/v05/data/labels_image_example.json"
) as f:
    attrs = json.load(f)

test_group.attrs.put(attrs)
test_group.create_array("0", shape=(1, 1, 1, 1, 1), dtype="uint8")
test_group.create_array("1", shape=(1, 1, 1, 1, 1), dtype="uint8")
test_group.create_array("2", shape=(1, 1, 1, 1, 1), dtype="uint8")

Image.from_zarr(image_group).labels
```
Traceback
```
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[1], [line 32](vscode-notebook-cell:?execution_count=1&line=32)
     29 test_group.create_array("1", shape=(1, 1, 1, 1, 1), dtype="uint8")
     30 test_group.create_array("2", shape=(1, 1, 1, 1, 1), dtype="uint8")
---> [32](vscode-notebook-cell:?execution_count=1&line=32) Image.from_zarr(image_group).labels

File ~/Projects/ome-zarr-models-py/src/ome_zarr_models/v05/image.py:237, in Image.labels(self)
    233     return None
    235 labels_group = self.members["labels"]
--> [237](https://file+.vscode-resource.vscode-cdn.net/Users/locerr/Projects/ome-zarr-models-py/~/Projects/ome-zarr-models-py/src/ome_zarr_models/v05/image.py:237) return Labels(attributes=labels_group.attributes, members=labels_group.members)

[...]

PR #239, opened 2025-08-26
Support unlistable stores in Labels

dstansby:
This adds support for unlistable stores to Labels. Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/236.


dstansby:
Ah, this is the wrong approach (at least for 0.4) because paths in the Labels group do not actually have to point to valid ImageLabels groups.

PR #243, opened 2025-08-28
Make v04 labels work with unlistable stores

dstansby:


dstansby:
pre-commit.ci autofix

PR #262, opened 2025-09-01
Add better instructions on implementing RFCs

dstansby:
I would definitely like some other folks to review this to see if it makes sense! Perhaps @LucaMarconato?


lorenzocerrone:
Sound clear to me too!

PR #280, opened 2025-09-13
feat: make zarr dependency optional

### LINKED ISSUE #161: Question re: dependency on zarr
I'm wondering about the use case where someone is writing a zarr file using something other than zarr-python, and they would like to comply with the ome-ngff spec.  I was hoping that using this package would let me ensure that my json files are valid and simplify construction of those objects.  However, I'd rather not depend on zarr-python in my use case.  Is a pure-model package that does not depend on zarr likely out of scope here?

tlambert03:
bare minimum required to make zarr dependency optional .

Not necessarily saying this is the best/cleanest way to do this, but it at least marks all the places where imports were made just for the sake of type hinting.

(closes #161)

testing this without zarr (i.e. running `pip uninstall zarr && pytest` and having it work) depends on https://github.com/tlambert03/pydantic-zarr/tree/drop-zarr (https://github.com/zarr-developers/pydantic-zarr/pull/112)


tlambert03:
oops sorry... weird version got in there.  one more time please? 😬

docs/index.md
```diff
@@ -11,6 +11,7 @@ The core of this package is a set of classes for representing different OME-Zarr
 | [`Labels`][ome_zarr_models.v05.Labels]         | [`Labels`][ome_zarr_models.v04.Labels]         |
 | [`ImageLabel`][ome_zarr_models.v05.ImageLabel] | [`ImageLabel`][ome_zarr_models.v04.ImageLabel] |
 | [`Well`][ome_zarr_models.v05.Well]             | [`Well`][ome_zarr_models.v04.Well]             |
+| [`Plate`][ome_zarr_models.v05.Plate]           | [`Plate`][ome_zarr_models.v04.Plate]           |
```
dstansby, line 14: I'm going to take a look at the `pydantic-zarr` PR before this one, but this looks like an oversight/bug 🤦 - would you be up for putting this fix in it's own PR?

tlambert03:
k, this is working again, and has pinned itself to the new main on pydantic-zarr, and fixes all the issues that came with https://github.com/zarr-developers/pydantic-zarr/pull/115

PR #281, opened 2025-09-15
feat: Add CLI validate/info

tlambert03:
this adds a new `ome-zarr-models` cli, with two (related) commands:

```bash
# simply validate and exit with proper return code.
ome-zarr-models validate https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0066/ExpD_chicken_embryo_MIP.ome.zarr

# print info
ome-zarr-models info https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0066/ExpD_chicken_embryo_MIP.ome.zarr
```

not doing anything fancy yet, just making sure the models get constructed properly.  No new dependencies.

relevant forum topic: https://forum.image.sc/t/where-to-place-ome-zarr-file-to-test-the-ngff-validator/104637/20?u=talley


tests/test_cli.py
```diff
@@ -0,0 +1,24 @@
...
+@pytest.mark.parametrize(
+    "url",
+    [
+        "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0066/ExpD_chicken_embryo_MIP.ome.zarr",
+        "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr",
+        "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0010/76-45.ome.zarr",
+    ],
+)
+@pytest.mark.parametrize("cmd", ["validate", "info"])
+def test_cli(url: str, cmd: str, monkeypatch: pytest.MonkeyPatch) -> None:
```
tlambert03, line 15: your call on whether you want to include this test.  can always be removed if those URLS become invalid (and currently the test will only xfail if the URL is invalid)

src/ome_zarr_models/v05/labels.py
```diff
@@ -80,6 +80,7 @@ class LabelsAttrs(BaseOMEAttrs):
     Attributes for an OME-Zarr labels dataset.
     """
 
+    version: Literal["0.5"] = "0.5"  # https://github.com/ome/ngff/pull/324
```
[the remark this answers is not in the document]
dstansby: Our policy here has always been that the specification is the text (not the JSON schema), and we will strictly implement the specification. We are aware that some data out there may not have the version, in which case it is not valid OME-Zarr data.

I understand this is practically a pain, but this is an issue caused by either the writers of that data, or the OME-Zarr specification editors, and it should be resolved at a specification (not an implementation level).

So this field should not be added here, unless we're currently interpreting the text of the 0.5 specification wrong somehow.

tests/test_cli.py
```diff
@@ -0,0 +1,24 @@
...
+    """Test the CLI commands."""
+    pytest.importorskip("zarr")
```
dstansby, line 17:
```suggestion
```
`zarr` is currently a dependency, so no need for this.

d-v-b:
> (it's surprisingly hard to find an actual valid v0.5 example out there 😂 )

That's what this library is designed to fix!

src/ome_zarr_models/__init__.py
```diff
@@ -77,16 +77,16 @@ def open_ome_zarr(group: zarr.Group) -> BaseGroup:
...
 
     raise RuntimeError(
-        f"Could not successfully validate {group} with any OME-Zarr group models.\n"
+        f"Could not successfully validate {group} against any OME-Zarr group model.\n"
         "\n"
-        "If you know what type of group you are trying to open, using the "
-        "<group class>.from_zarr() method will give you a more informative "
-        "error message explaining why validation failed."
+        "The following errors were encountered while trying to validate:\n\n"
+        + "\n\n".join(f"- {type(e).__name__}: {e}" for e in errors)
```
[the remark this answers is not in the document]
dstansby: Ooh okay, that string of messages is much better/shorter than I was expecting (in general I always expect the worse with pydantic errors...). Happy to keep your change in that case 👍

d-v-b:
>  One approach might be to warn on confusions or instabilities, and then cast to your current interpretation of what valid actually means.

That's an interesting idea, we could basically take a "parse, don't validate" approach. This would widen the types of data that this library consumes, but the output of this library would still be strictly typed. And users could be notified when input had to be coerced / corrected. Given the inconsistent implementations out there, this is probably a good direction from a "use this library with the most data" perspective.

d-v-b:
the proposal here is equivalent to defining _two types_: an input type and an output type, where the input type is wider than the output type, along with a well-behaved (non-invertible) transformation from instances of the input type to instances of the output type (i.e., parsing).

dstansby:
Ooh, I hadn't come across Postel's law before, or considered the difference between reading and writing really. This is definitley off topic now though - could you open an fresh issue (can just be a copy/paste of some of the comments above) about being less strict with input data where the meaning would be unambiguous.

src/ome_zarr_models/cli.py
```diff
@@ -0,0 +1,91 @@
...
+        action="version",
+        version=f"ome-zarr-models version {__version__}",
+    )
+    subparsers = parser.add_subparsers(
+        dest="command", help="Available commands", metavar="COMMAND"
+    )
+
+    # validate sub-command
+    validate_cmd = subparsers.add_parser("validate", help="Validate an OME-Zarr")
+    validate_cmd.add_argument("path", type=str, help="Path to zarr to validate")
```
dstansby, line 24:
```suggestion
    validate_cmd.add_argument("path", type=str, help="Path to OME-Zarr group to validate")
```

src/ome_zarr_models/cli.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # info sub-command
+    info_cmd = subparsers.add_parser("info", help="Get information about an OME-Zarr")
```
dstansby, line 27:
```suggestion
    info_cmd = subparsers.add_parser("info", help="Get information about an OME-Zarr group")
```

src/ome_zarr_models/cli.py
```diff
@@ -0,0 +1,91 @@
...
+    info_cmd.add_argument(
+        "path", type=str, help="Path to zarr to get information about"
```
dstansby, line 29:
```suggestion
        "path", type=str, help="Path to OME-Zarr group to get information about"
```

src/ome_zarr_models/cli.py
```diff
@@ -0,0 +1,91 @@
...
+        case "validate":
+            validate(args.path)
+        case "info":
+            info(args.path)
+        case None:
+            parser.print_help()
+            sys.exit(1)
+
+
+def validate(path: str) -> None:
```
dstansby, line 45: It would be nice to add the option to valdate against a specific version and group-type. Something like:

```python
def validate(
    path: str,
    version: Literal["0.4", "0.5"] | None = None,
    group_type: Literal["image", "hcs", ...] | None = None
) -> None:
```

What do you think? Happy to punt that to a follow up PR, since it might need some more internal logic adding.

src/ome_zarr_models/cli.py
dstansby, line 1: Could you make this `_cli.py` instead to mark that everythig in it is considered private.

tests/test_cli.py
```diff
@@ -0,0 +1,92 @@
...
+    "version,json_fname",
+    [
+        ("0.4", "multiscales_example.json"),
+        ("0.5", "image_example.json"),
+        ("0.5", "image_label_example.json"),
+        ("0.5", "plate_example_1.json"),
+    ],
+)
+@pytest.mark.parametrize("cmd", ["validate", "info"])
+def test_cli(
```
dstansby, line 78: Would it be easy to add a test against a non-valid OME-Zarr dataset too (to make sure an error is raised/the right message is outputted)?

tests/test_cli.py
```diff
@@ -0,0 +1,92 @@
...
+    tmp_path: Path,
+) -> None:
+    """Test the CLI commands."""
+
+    zarr_group = json_to_zarr_group(
+        version=version, json_fname=json_fname, store=LocalStore(root=tmp_path)
+    )
+    populate_fake_data(zarr_group)
+    monkeypatch.setattr("sys.argv", ["ome-zarr-models", cmd, str(tmp_path)])
+    main()
```
dstansby, line 92: Would be good to check stdout after here to make sure there's a success message.

dstansby (changes requested):
Looks great overall - a few requests for changes, and a couple of questions.

tlambert03:
all set!

d-v-b:
this looks great @tlambert03!

Issue #283, opened 2025-09-16
Add user-facing docs for new CLI


dstansby:
Before https://github.com/ome-zarr-models/ome-zarr-models-py/pull/281 goes into a release, some docs should be added so users know the fancy new CLI exists.

PR #285, opened 2025-09-19
Add docs for CLI

dstansby:
Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/283


tlambert03 (approved):
looks good!  is `info.txt` still required?

dstansby:
>  is info.txt still required?

🙈 woops, thanks for catching! It is not

Issue #294, opened 2025-09-30
`Axis.name` should be required


LDeakin:
This implementation has `name` as optional in the `axes` metadata.

https://github.com/ome-zarr-models/ome-zarr-models-py/blob/6c51997beda0fb94fb271fab7ebd38ba37ba2f60/src/ome_zarr_models/common/axes.py#L19-L22

The spec says it is required https://ngff.openmicroscopy.org/0.5/index.html#axes-md

> "axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes a dimension (axis) and:
>    MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.

Is it feasible to align this to the spec? Validation against `dimension_names` in Zarr metadata seems orthogonal.

PR #295, opened 2025-09-30
Make name required in axes

dstansby:
Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/294.

I also took the oppurtunity to remove the common code for axes and multiscales, so that each version has it's own copy. While this duplicates code, I think versions of the specification diverge enough that having different copies makes how the objects are constructed more obvious (less redirection for developers to understand one of the metadata classes full structure). There might be scope to de-duplicate some of the validation logic within the classes later.


LDeakin:
Just flagging that the type hint still has `str | None`

https://github.com/ome-zarr-models/ome-zarr-models-py/blob/408b09d14f580c1bf3bbc1f3dc202e2fcfca54fe/src/ome_zarr_models/v05/axes.py#L22

dstansby:
> Just flagging that the type hint still has str | None

Yes, this is intentional, because the specification does not specify that that `name` has to be a string.

d-v-b:
there's an open issue over in the spec repo: https://github.com/ome/ngff/issues/326

Issue #293, opened 2025-09-26
Test against minimum supported dependencies


dstansby:
As suggested in https://github.com/ome-zarr-models/ome-zarr-models-py/issues/161#issuecomment-3336027290; this should be easy with the `uv sync --resolution=lowest-direct` command to setup a venv with uv.

PR #304, opened 2025-10-01
Test with minimum dependencies

dstansby:
Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/293.


d-v-b (approved):

dstansby:
@d-v-b could you take another look at this? I had the idea to pin exactly the development dependencies, and then use dependabot to update them periodically. This way (especially for the docs) we're always building in a known working environment, and can check that updated dependencies don't break any development before updating them.

Issue #270, opened 2025-09-04
Test-driven development for RFC-5


LucaMarconato:
RFC-5 support introduces several tests for data form external sources (as of today 68 in total). Currently only 1 is implemented and 67 are skipped.
They are located in these files. We should work towards fixing them.

`tests/_v06/test_data_rfc5/ngff/`
- [ ] test_coordsystems.py
- [ ] test_multiscales_strict.py
- [ ] test_subspace.py
- [ ] test_transformations.py

`tests/_v06/test_data_rfc5/ngff_rfc5_coordinate_tranformation_examples/`
- [ ] test_2d_axis_dependent.py
- [ ] test_2d_basic.py
- [ ] test_2d_basic_binary.py
- [ ] test_2d_nonlinear.py
- [ ] test_2d_simple.py
- [ ] test_3d_axis_dependent.py
- [ ] test_3d_basic.py
- [ ] test_3d_basic_binary.py
- [ ] test_3d_nonlinear.py
- [ ] test_3d_simple.py
- [ ] test_user_stories.py

dstansby:
These are now covered by newer tests, so I'll close this as done.

PR #305, opened 2025-10-01
Initial RFC 5 implementation

dstansby:
This is a very manual cherry-pick of all the comments of the old `rfc5` branch, which I have now renamed to `rfc5-old`. I tried doing a proper rebase, but it was proving too hard. In addition, I have only picked commits up to the point just before it was merged into the `_v06` submodule. I would like to do this move myself again to understand the code, and maybe make some improvements/simplifications along the way.

I plan to review this PR myself, perhaps make some improvements/changes, and then merge the code into the `_v06` submodule, before merging this PR into the new `rfc5` branch.

cc @LucaMarconato , @melonora - I hope this is okay. The commit history and attribution should remain.


LucaMarconato:
Hi @dstansby, fine by me. I will be busy with a deadline for the next 2 weeks, so if this makes the implementation faster, that's great!

> In order to speed up finishing the implementation

I don't fully agree with this: it is true that the implementation is faster because the old approach was a test-driven development with 67 tests https://github.com/ome-zarr-models/ome-zarr-models-py/issues/270, which are a lot to implement and this takes time. On the other hand, I still think that in one way or the other (i.e. using GitHub submodule vs downloading the data via code vs by copy-pasting the data into this repo), we should test against all the datasets made available in https://github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples from @bogovicj.

Not doing this means duplicating the effort to test all the edge cases in a way that is not easily reusable across programming languages. And the datasets above would benefit from discovering fixes together, as @will-moore did here: https://github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aclosed.

In the old code from my PR I was also adding tests against the examples here: https://github.com/bogovicj/ngff/tree/dev/examples. I would actually skip these examples. Even if they should be more "official", since they would appear in the NGFF repo, they were not examples of complete objects (i.e. transformations inside multiscale, datasets, etc.). So, here we agree that removing this second sets of tests is a good way to speed up development without sacrificing much coverage of edge cases.

LucaMarconato:
Anyway, I'll keep following the progress and I'll be ready to do a push before the NGFF hackathon as soon as I get free from the other coding duties.

PR #308, opened 2025-10-02
Minor changes to coordinate transform object names

dstansby:
I generally find (at least here in ome-zarr-models, and also in pydantic-models) that a collection of specific objects is called `Any...`, so I renamed CoordinateTransformation -> AnyCoordinateTransformation, which then frees up CoordinateTransformation to be used for the base class. I think this is a bit nicer.


d-v-b:
as long as it doesn't confuse people, I think this would be a big readability improvement

PR #309, opened 2025-10-02
Add back RFC5 example tests

dstansby:
This adds John's RFC5 examples back to the tests, with a slightly different way of testing them. The directory tree is automatically searched for all .zarr files, and then these are attempted to be loaded with `Image.from_zarr()`. This is less complex than the previous implementation, and also explicitly lists the files that are currently broken, making it easy to see which tests we still need to fix.

@LucaMarconato if you have time to look at this it would be appreciated. I don't think any of the tests were actually implemented for these examples in your original PR, so I'm not sure if your intention was for the test to be what I've implemented here (using `Image.from_zarr()`), or something else?


tests/_v06/rfc5/transform_examples/test_transform_examples.py
```diff
@@ -0,0 +1,201 @@
+from pathlib import Path
+
+import pytest
+import zarr
+
+from ome_zarr_models._v06.image import Image
+
+from .expected_attributes import expected_attrs
```
LucaMarconato, line 8: Code philosophy comment: I usually like to have full paths instead of relative ones for imports. Such as from `tests._v06.rfc5.transform_examples.expected_attributes`. 
VSCode (I'm using the GitHub webapp) wasn't able to resolve "Go To References" for `expected_attrs`. I wonder if it's because of the use of relative imports.
Any preference for the relative one, beside brevity?

dstansby: No reason, just being lazy - thanks for catching

PR #314, opened 2025-10-04
Add code to transform points

dstansby:
This adds a `transform_point` method to `Transform`, and implements it for most transforms. Through doing this I've found some implementation issues that should be raised with RFC5:

1. `mapAxis` does not operate on coordinates without axes names like the rest of the transforms. Instead the coordinates need axis names attached to apply the transform.
2. `byDimension` has the same problem as ☝️ , and also specifies `input`/`output` as a list of Axes, which is different to every other transform which specifies coordinate system names.
3. `inverseOf` seems to be redundant - why can one not just manually define an inverse transform that has flipped input/output coordinate systems? If there is a use case and it is not redundant, it is at least very confusing to me why this exists.
4. The metdata in https://ngff.openmicroscopy.org/rfc/5/index.html#coordinatetransformations-metadata for `rotation` has the wrong type (it should be `list[list[float]]`, not `list[float]`)

### DIFF
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -463,9 +463,6 @@ class ByDimension(Transform):
     def transform_point(self, point: typing.Sequence[float]) -> tuple[float, ...]:
         raise NotImplementedError
 
-    def inverse_transform_point(self, point: typing.Sequence[float]) -> TPoint:
-        raise NotImplementedError
-
 
 AnyTransform = Annotated[
     Identity


dstansby:
I think there is high value in having the actual transforms implemented in one core package like this, to reduce duplication and reduce the number of packges users have to install to use the transformations (and hence confusion when setting up an environment). We do have to balance that with not letting `ome-zarr-models` become too large or grow extra dependencies. When it comes to transforming points I think the benefits of having a centralised implementation are worth the costs of extra code here.

> but for applications on real data I'd have the users rather relying on another library (the "ngff-transformations" library that we brainstormed at the NGFF hackathon.

Do you have other reasons for the preference to depend on another library?

> I could imagine ome-zarr-models-py to split the transform_points() function into some parts, some functions to perform validation on the input data and matrix, to be overridden by ngff-transformations or other implementations, and then the actual code to transform the data. What do you think?

I'm not sure what the motivation would be to split the validation and transform code into different parts, and not sure what this would look like in practice? What would be the motivation for doing this?

> The rationale is that transform points assumes the data to be in a particular canonical format (e.g. dataframe, geodataframe, numpy array, Xarray object, or dask object, cupy.array, torch.tensor) are all good choices.

Good point - for simplicity I just wrote the current code to accept Python builtin objects. Is there a common API that all of these other array-like objects share, that we could possibly try and support here? I've heard about the "Array API" before, but not sure how widely it's used/supported.

As written, I don't think there's anything stopping other libraries sub-classing the `Transform` classes and adding their own methods to transform other objects (e.g., dataframes, torch arrays), so this approach still maintains extensibility.

dstansby:
@LucaMarconato I had a bit more of a think, and I think implementing `transform_point` here is nice, as it can serve as a reference implementation for other libraries to build upon. I will add some docs to that effect, encouraging other libraries to add their own transform methods for different data types.

dstansby:
That's a nice idea! Could you open a separate issue so we don't forget?

LucaMarconato:
Done: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/363

PR #327, opened 2025-10-16
Add basic container for loading RFC-5 parent groups

dstansby:
This adds a new `Container` class to handle the case of the new type of group in RFC 5 that stores "child" images, and coordinate transforms alongside. At the moment I've just implemented the logic for reading the metadata, but I can imaging there's use cases for adding extra functionality later down the line.

I'm not sure if `Container` is a good name; **other suggestions encouraged**...!

In order to know which paths to look for Image groups, I had to modify the example so the "input" coordinate systems that point to a path start with `/`. Distinguishing coordinate system names from coordinate system paths may change at some point after discussion between myself, @jo-mueller and @will-moore in github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples/issues/11.

### DIFF
--- a/src/ome_zarr_models/_v06/collection.py
+++ b/src/ome_zarr_models/_v06/collection.py
@@ -12,7 +12,7 @@ from ome_zarr_models._v06.coordinate_transforms import (
 from ome_zarr_models._v06.image import Image
 
 
-class ContainerAttrs(BaseOMEAttrs):
+class CollectionAttrs(BaseOMEAttrs):
     coordinateTransformations: tuple[AnyTransform, ...] = Field(default=())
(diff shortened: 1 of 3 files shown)


jo-mueller:
> I'm not sure if Container is a good name; other suggestions encouraged...!

I was also wondering about that question. I was thinking about "Spatial collection" or "Transform(ation) set"? There's a good chance that the transformations parent-level group will later be absorbed by rfc6 and move into the `attributes` section of the newly introduced collections.

will-moore:
I added a `CoordinateSystems` class as the "container" at https://github.com/will-moore/napari-ome-zarr/pull/2/files#diff-1c9c253d036df88195dd4242fc3afde18a2f409c5e98e622b9421a823ac6dac2R269

PR #328, opened 2025-10-22
Add example of getting transform graph of a collection

dstansby:
This adds a rough example of constructing a transform graph for one of the user examples:

image

In general it worked quite well - the only think I had to change with the metadata was making sure paths started with "/". We're discussing how to distinguish paths and named coordinate systems in the thread here: https://github.com/bogovicj/ngff-rfc5-coordinate-transformation-examples/issues/11


dstansby:
Superseeded/rewritten by https://github.com/ome-zarr-models/ome-zarr-models-py/pull/331

PR #347, opened 2025-11-14
Add Collection.new() method

fstur:


src/ome_zarr_models/_v06/collection.py
```diff
@@ -50,6 +51,53 @@ def from_zarr(cls, group: zarr.Group) -> Self:  # type: ignore[override]
         """
         return _from_zarr_v3(group, cls, CollectionAttrs)
 
+    @classmethod
+    def new(
+        cls,
+        *,
+        images: Sequence[tuple[str, Image]],
```
dstansby: If the type here is changed to `dict[str, Image]`, it would ensure no duplicate paths (because dictionary keys can't be duplicated)

PR #378, opened 2026-01-03
Test on Python 3.14

dstansby:
Closes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/400


dstansby:
I'm going to investigate this more thoroughly at https://github.com/ome-zarr-models/ome-zarr-models-py/issues/400

PR #415, opened 2026-04-01
Relax well image constraint

rzlim08:
* The spec for images in wells has been relaxed from alphanumeric to alphanumeric + ".-_" as detailed [here](https://github.com/ome/ngff-spec/pull/71/changes)


dstansby (dismissed):
Thanks a lot for the PR! This overall looks good to me. A few points that should be improved:

- We can't change the behaviour of version 0.5, just the upcoming 0.6. So i think you will need to move `WellImage` out of `ome_zarr_models/common/well_types.py`, and make a copy of it in each version's files (0.4, 0.5, and 0.6), and then only edit the on in 0.6. I think then the new `WellImageConstraint` should also live somewhere in `ome_zarr_modles/_v06`, since it's not really a generic constraint but specific to OME-Zarr 0.6 (and onwards).
- It would be nice to add to `common/validation.py` a new `validate_zarr_node`, that encodes the logic given in https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/core/index.rst#node-names, and then have the allowed characters (the `[A-Za-z0-9_.-]` bit) as separate logic that's OME-Zarr well specific.

Thanks again for the contribution

rzlim08:
> Thanks a lot for the PR! This overall looks good to me. A few points that should be improved:
> 
> * We can't change the behaviour of version 0.5, just the upcoming 0.6. So i think you will need to move `WellImage` out of `ome_zarr_models/common/well_types.py`, and make a copy of it in each version's files (0.4, 0.5, and 0.6), and then only edit the on in 0.6. I think then the new `WellImageConstraint` should also live somewhere in `ome_zarr_modles/_v06`, since it's not really a generic constraint but specific to OME-Zarr 0.6 (and onwards).
> * It would be nice to add to `common/validation.py` a new `validate_zarr_node`, that encodes the logic given in https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/core/index.rst#node-names, and then have the allowed characters (the `[A-Za-z0-9_.-]` bit) as separate logic that's OME-Zarr well specific.
> Thanks again for the contribution

Thanks, thoughts on this approach? I've kept the WellImageConstraint in the "common" folder and made a custom `WellImage` for versions 0.4 and 0.5. This will allow future versions to inherit the current logic. 

If it makes more sense I can move it into 0.6 though. 

Thanks for the response!

src/ome_zarr_models/common/well_types.py
```diff
@@ -21,7 +21,7 @@ class WellImage(BaseAttrs):
     A single image within a well.
     """
 
-    path: Annotated[str, AlphaNumericConstraint]
+    path: Annotated[str, AfterValidator(validate_zarr_node)]
```
dstansby: As well as checking this is a valid Zarr node, I think that there needs to be some additional validation to make sure that 

> The `path` MUST only use characters in the sets `a-z`, `A-Z`, `0-9`, `-`, `_`, `.`. 

Could you add an additional constraint (maybe called `WellPathNameConstraint`?) that is a copy of `AlphaNumericConstraint`, but also allows for the -_. characters?

rzlim08: I've updated this to become `WellPathCharsConstraint`. Sorry for the delay in update.

jo-mueller (approved):

jo-mueller:
@rzlim08 looks good to me! @dstansby won't be able to maintain this project in his official capacity anymore and I'll try my best to take over and ensure continuity. *However*, since we both have equal rights, my PR approval doesn't supersede David's 🙈 I think the easiest way to get this merged would be to close and re-open the PR and then I can merge it.

dstansby:
@jo-mueller I think you could manually dismiss my review?

rzlim08:
thanks @jo-mueller. let me know if you can't dismiss the review

jo-mueller:
> @jo-mueller I think you could manually dismiss my review?

Thanks, didn't know that was possible 👍

PR #416, opened 2026-04-13
Allow the colors field to be absent

rzlim08:
* image-label > colors is not a required field accorrding to [NGFF spec](https://ngff.openmicroscopy.org/specifications/dev/schemas/label.html#property-label-ome-image-label-colors)
* The pydantic model already accepts `None` as a value, however the field validator assumes it's a list. 
* Pins `pydantic` version to `<3.13`. Tests fail with `Class not fully defined ` errors otherwise

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ dynamic = ["version"]
 description = "A minimal Python package for reading OME-Zarr (meta)data "
 readme = "README.md"
 requires-python = ">=3.11, <3.14"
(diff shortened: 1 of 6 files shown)


dstansby (commented):
Thanks for catching and fixing this! Can you add a fix to the same code in _v06 too?

rzlim08 (commented):
No problem! Added a test for v0.6

dstansby (approved):

PR #418, opened 2026-04-15
Changelog for 1.7

dstansby:
@rzlim08 since you made these changes, could you take a quick look and see if this is okay?


rzlim08 (approved):
The only thing to note is that the color change should apply across all versions. Otherwise looks good

PR #420, opened 2026-04-23
Finish multiscale validation for 0.6

dstansby:


dstansby:
@jo-mueller could you review this? I think two examples need fixing to pass the new validation, which I'll open a PR for in the examples repo.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -240,23 +209,31 @@ def _ensure_axes_dataset_transforms(data: Self) -> Self:
...
+        Make sure datasets are ordered from the highest resolution to smallest.
         """
         scale_transforms: list[Scale] = []
         for dataset in datasets:
             (transform,) = dataset.coordinateTransformations
             if isinstance(transform, Scale):
                 scale_transforms.append(transform)
+            elif isinstance(transform, Identity):
+                # TODO: is there some way we can compare this to other scales or
+                #  identities?
```
jo-mueller, line 228: My interpretation of using `Identity` in the multiscales > datasets > coordinateTransformations, would be that `Identity` is equivalent to `Scale(scale=(1, 1,1)` for as many dimensions as there are. That means, that an `Identity` should really only ever occur in the first dataset.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -376,6 +352,65 @@ def check_unique_system_names(
...
+    @model_validator(mode="after")
+    def _ensure_axis_types(self) -> Self:
+        """
+        Ensures that the following conditions are true:
+
+        - there are only 2 or 3 axes with type `space`
+        - the axes with type `space` are last in the list of axes
+        - there is only 1 axis with type `time`
+        - there is only 1 axis with type `channel`
+        - there is only 1 axis with a type that is not `space`, `time`, or `channel`
```
jo-mueller, line 376: Not sure why two custom axes shouldn't be allowed?

tests/_v06/test_multiscales.py
```diff
@@ -365,8 +369,8 @@ def test_from_v05() -> None:
                 coordinateTransformations=(
                     Sequence(
                         type="sequence",
-                        input=CoordinateSystemIdentifier(path="0"),
-                        output=CoordinateSystemIdentifier(name="intrinsic"),
+                        input=CoordinateSystemIdentifier(name=None, path="0"),
+                        output=CoordinateSystemIdentifier(name="intrinsic", path=None),
```
jo-mueller, line 373: No objections in general here, but don't `path` and `name` always default to `None`? Is there a specific reason to make it explicit?

jo-mueller (approved):

dstansby:
Here's the corresponding PR to fix the examples: https://github.com/jo-mueller/ngff-rfc5-coordinate-transformation-examples/pull/43

src/ome_zarr_models/_v06/multiscales.py
dstansby: [RFC 5 says](https://ngff.openmicroscopy.org/rfc/5/index.html#multiscales-metadata):

> MAY contain one additional entry of “type:channel” or a null / custom type

jo-mueller:
@dstansby beat me to it 😆

src/ome_zarr_models/_v06/multiscales.py
dstansby: You could have scales that are greater than 1 though, which means an identity could appear in the middle of the list.

The tricky thing is that identity doesn't declare its dimensionality, so when we only see an identity transform we need to construct a `Scale(1, ...)` object to compare, but don't know what the dimensionality of that scale object should be.

Is allowing an identity transfrom here new to RFC5 recently? It complicates things slightly, and I think it would be nicer to not allow it.

jo-mueller:
Also - does this branch need a rebase? I feel like some of these changes were already on #428 🤔

tests/_v06/test_multiscales.py
dstansby: 👍 no reason, I just copy/pasted pytest output. I'll remove them.

dstansby:
> Also - does this branch need a rebase?

I don't think so... which bits do you think are duplicated?

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: Uh, I think that statement somehow didn't make it into the current spec!

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: No objections. It would be a minor change and these *meh* effects are exactly what the spec phase is for.

jo-mueller: Corresponding issue: https://github.com/ome/ngff/issues/513

tests/_v06/test_multiscales.py
```diff
@@ -303,8 +303,8 @@ def test_default_coordinate_systems() -> None:
     assert multiscale.intrinsic_coordinate_system == CoordinateSystem(
         name="physical",
         axes=(
-            Axis(name="j", type=None, discrete=None, unit=None, longName=None),
-            Axis(name="i", type=None, discrete=None, unit=None, longName=None),
+            Axis(name="j", type="space", discrete=None, unit=None, longName=None),
+            Axis(name="i", type="space", discrete=None, unit=None, longName=None),
```
jo-mueller, line 307: I *thought* I had this changed, too. But maybe I am confusing branches :) Nevermind, it's correct after all 👍

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: Wait, nvm, found it. Looks good then!

tests/_v06/test_multiscales.py
jo-mueller: Nevermind, that was on #432

PR #419, opened 2026-04-22
Unify input output

jo-mueller:
This PR reflects the changes in https://github.com/ome/ngff-spec/pull/117

I think right now the largest contribution to failing tests are the remote testing data, that has yet to be updated to 0.6.dev4. Relevant working heads:

- https://github.com/ome/ngff-spec/pull/117
- https://github.com/jo-mueller/ngff-rfc5-coordinate-transformation-examples/pull/38

PR #425, opened 2026-04-23
feat: make interpolation optional

jo-mueller:
Title. The interpolation field of the `coordinates` and `displacements` transform should be optional.

### DIFF
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -616,6 +616,9 @@ class Displacements(Transform):
 
     type: Literal["displacements"] = "displacements"
     path: str = Field(..., description="Path to the Zarr array displacement field.")
+    interpolation: str | None = Field(
+        default=None, description="Interpolation method to be used when applying the transform (default: 'linear')."
+    )
 
     @property
     def has_inverse(self) -> bool:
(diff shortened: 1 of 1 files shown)


src/ome_zarr_models/_v06/coordinate_transforms.py
```diff
@@ -616,6 +616,9 @@ class Displacements(Transform):
 
     type: Literal["displacements"] = "displacements"
     path: str = Field(..., description="Path to the Zarr array displacement field.")
+    interpolation: str | None = Field(
+        default=None, description="Interpolation method to be used when applying the transform (default: 'linear')."
```
dstansby:
```suggestion
        default=None, description="Interpolation method to be used after applying the transform."
```

src/ome_zarr_models/_v06/coordinate_transforms.py
```diff
@@ -645,6 +648,11 @@ class Coordinates(Transform):
     path: str = Field(
         ..., description="Path to the Zarr array containing the coordinate mapping."
     )
+    interpolation: str | None = Field(
+        default=None,
+        description="Interpolation scheme that should be used when applying"
+        " the transform.",
```
dstansby:
```suggestion
        description="Interpolation method to be used after applying the transform."
```

dstansby (commented):
- I don't think the default should be mentioned, or if it should be it needs rephrasing. I can see cases where you want to default to either linear (images) or nearest neighbour (labels)
- The interpolation is applied after the transform, I clarified the strings to that effect.

What do you think?

jo-mueller:
> I don't think the default should be mentioned

Agree. Right now I'm just spawning PRs from #419 so some changes I had already made may be lost in the process 😬 Thanks for the extra pair of eyes!

dstansby:
No problem, and thanks for splitting it out into lots of PRs, makes it much easier to review and merge :)

dstansby (approved):

dstansby:
I'll ignore pre-commit for now and we can fix it in one go later

PR #426, opened 2026-04-23
feat: update ByDimension transform

jo-mueller:
Structure for `byDimension` transform needs another field to express how transforms are wrapped specifically in 0.6.dev4

### DIFF
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -714,8 +714,8 @@ class ByDimensionTransform(BaseAttrs):
     """
     A transformation item within a byDimension coordinate transformation.
 
-    Per RFC-5, each transformation within a byDimension must specify which
-    axes it operates on via input_axes and output_axes.
+    As well as the actual transform, this specifies which
+    axes it operates on via the input_axes and output_axes fields.
     """
 
     transformation: "AnyTransform" = Field(..., description="The coordinate transformation.")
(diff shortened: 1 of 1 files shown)


src/ome_zarr_models/_v06/coordinate_transforms.py
```diff
@@ -702,10 +702,17 @@ def as_affine(self) -> "Affine":
             ) from e
 
 
-class ByDimensionTransform(BaseAttrs):
-    transformation: "AnyTransform" = Field(..., description="Transformation.")
-    input_axes: tuple[int, ...] = Field(..., description="Input axis indices.")
-    output_axes: tuple[int, ...] = Field(..., description="Output axis indices.")
+class ByDimensionTransformation(BaseAttrs):
```
dstansby: I'd rather stick with `Transform` instead of `Transformation` if that's okay? It's what we used for the base class, and saves a few characters.

src/ome_zarr_models/_v06/coordinate_transforms.py
```diff
@@ -702,10 +702,17 @@ def as_affine(self) -> "Affine":
...
+    """
+    A transformation item within a byDimension coordinate transformation.
+
+    Per RFC-5, each transformation within a byDimension must specify which
+    axes it operates on via input_axes and output_axes.
```
dstansby:
```suggestion
    As well as the actual transform, this specifies which
    axes it operates on via the input_axes and output_axes fields.
```

src/ome_zarr_models/_v06/coordinate_transforms.py
jo-mueller: No objections!

dstansby (approved):

PR #427, opened 2026-04-23
feat: remove default coordinate system

jo-mueller:
Used to be declared as default because of its position in the coordinate systems list. This ordering is no longer relevant so the default coordinate system is no longer necessary.


dstansby (changes requested):
Looks like a test needs fixing - at a glance it probably just needs deleting??

jo-mueller:
Jup! Beat me to it :)

dstansby (approved):

PR #424, opened 2026-04-23
Update transformation examples submodule

jo-mueller:
Breakout from #422 . Bumps rfc5 transform example submodule to latest version

PR #428, opened 2026-04-23
Cs identifier everywhere

jo-mueller:
Biggest chunk of work from #419 

This makes consistent use of the `CoordinateSystemIdentifier` throughout all circumstances and updates the corresponding validation checks and tests.

Shouldbe merged only after #424

### DIFF
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -70,14 +70,14 @@ class CoordinateSystemIdentifier(BaseAttrs):
     relative to the current Zarr group.
     """
 
-    name: str = Field(..., description="Coordinate system name.")
+    name: str | None = Field(
+        default=None,
+        description="Coordinate system name."
+        )
     path: str | None = Field(
         default=None,
-        description="Path to Zarr group where the coordinate system is defined.",
+        description="Path to Zarr group.",
     )
-
-
-TCoordSysIdentifier = CoordinateSystemIdentifier | str | None
 
 
 class Transform(BaseAttrs, ABC):
@@ -95,10 +95,10 @@ class Transform(BaseAttrs, ABC):
     """
 
     type: str = Field(..., description="Unique identifier for type of transform.")
-    input: TCoordSysIdentifier = Field(
+    input: CoordinateSystemIdentifier | None= Field(
         default=None, description="Input coordinate system identifier."
     )
-    output: TCoordSysIdentifier = Field(
+    output: CoordinateSystemIdentifier | None= Field(
         default=None, description="Output coordinate system identifier."
     )
     name: str | None = Field(
@@ -378,8 +378,8 @@ class Affine(Transform):
         cls,
         matrix: typing.Sequence[typing.Sequence[float]],
         vector: typing.Sequence[float],
-        input: TCoordSysIdentifier = None,
-        output: TCoordSysIdentifier = None,
+        input: CoordinateSystemIdentifier = None,
+        output: CoordinateSystemIdentifier = None,
         name: str | None = None,
     ) -> Self:
         return cls(
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -711,9 +711,18 @@ class Bijection(Transform):
 
 
 class ByDimensionTransform(BaseAttrs):
-    transformation: "AnyTransform" = Field(..., description="Transformation.")
-    input_axes: tuple[int, ...] = Field(..., description="Input axis indices.")
-    output_axes: tuple[int, ...] = Field(..., description="Output axis indices.")
+    """
+    A transformation item within a byDimension coordinate transformation.
+
+    As well as the actual transform, this specifies which
+    axes it operates on via the input_axes and output_axes fields.
+    """
+
+    transformation: "AnyTransform" = Field(
+        ..., description="The coordinate transformation."
+    )
+    input_axes: tuple[int, ...] = Field(..., description="Input axes indices.")
+    output_axes: tuple[int, ...] = Field(..., description="Output axes indices.")
 
 
 class ByDimension(Transform):
@@ -754,3 +763,6 @@ AnyTransform = Annotated[
     | ByDimension,
     Field(discriminator="type"),
 ]
+
+# Rebuild models to resolve forward references
+ByDimensionTransform.model_rebuild()
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -82,9 +82,7 @@ class CoordinateSystemIdentifier(BaseAttrs):
         Ensures that at least one of name or path is given.
         """
         if self.name is None and self.path is None:
-            raise ValueError(
-                "At least one of 'name' or 'path' must be set."
-                )
+            raise ValueError("At least one of 'name' or 'path' must be set.")
         return self
 
 
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -617,7 +617,8 @@ class Displacements(Transform):
     type: Literal["displacements"] = "displacements"
     path: str = Field(..., description="Path to the Zarr array displacement field.")
     interpolation: str | None = Field(
-        default=None, description="Interpolation method to be used after applying the transform."
+        default=None,
+        description="Interpolation method to be used after applying the transform.",
     )
 
     @property
@@ -650,7 +651,7 @@ class Coordinates(Transform):
     )
     interpolation: str | None = Field(
         default=None,
-        description="Interpolation method to be used after applying the transform."
+        description="Interpolation method to be used after applying the transform.",
     )
 
     @property
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -76,6 +76,17 @@ class CoordinateSystemIdentifier(BaseAttrs):
         description="Path to Zarr group.",
     )
 
+    @model_validator(mode="after")
+    def _ensure_either_name_or_path(self: Self) -> Self:
+        """
+        Ensures that at least one of name or path is given.
+        """
+        if self.name is None and self.path is None:
+            raise ValueError(
+                "At least one of 'name' or 'path' must be set."
+                )
+        return self
+
 
 class Transform(BaseAttrs, ABC):
     """
--- a/src/ome_zarr_models/_v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/_v06/coordinate_transforms.py
@@ -70,10 +70,7 @@ class CoordinateSystemIdentifier(BaseAttrs):
     relative to the current Zarr group.
     """
 
-    name: str | None = Field(
-        default=None,
-        description="Coordinate system name."
-        )
+    name: str | None = Field(default=None, description="Coordinate system name.")
     path: str | None = Field(
         default=None,
         description="Path to Zarr group.",
@@ -95,10 +92,10 @@ class Transform(BaseAttrs, ABC):
     """
 
     type: str = Field(..., description="Unique identifier for type of transform.")
-    input: CoordinateSystemIdentifier | None= Field(
+    input: CoordinateSystemIdentifier | None = Field(
         default=None, description="Input coordinate system identifier."
     )
-    output: CoordinateSystemIdentifier | None= Field(
+    output: CoordinateSystemIdentifier | None = Field(
         default=None, description="Output coordinate system identifier."
     )
     name: str | None = Field(
@@ -378,8 +375,8 @@ class Affine(Transform):
         cls,
         matrix: typing.Sequence[typing.Sequence[float]],
         vector: typing.Sequence[float],
-        input: CoordinateSystemIdentifier = None,
-        output: CoordinateSystemIdentifier = None,
+        input: CoordinateSystemIdentifier | None = None,
+        output: CoordinateSystemIdentifier | None = None,
         name: str | None = None,
     ) -> Self:
         return cls(
--- a/src/ome_zarr_models/_v06/multiscales.py
+++ b/src/ome_zarr_models/_v06/multiscales.py
@@ -72,7 +72,8 @@ class Multiscale(BaseAttrs):
         if output is None:
             raise ValueError(
                 "Output coordinate system in "
-                f"{self.datasets[0].coordinateTransformations[0]} not found. ")
+                f"{self.datasets[0].coordinateTransformations[0]} not found. "
+            )
         return next(cs for cs in self.coordinateSystems if cs.name == output.name)
 
     @classmethod
@@ -306,7 +307,6 @@ class Multiscale(BaseAttrs):
         # check: additional coordinate transformations must have `name` set
         # in both input AND output
         for transformation in self.coordinateTransformations:
-
             # first check that both exist
             if transformation.input is None:
                 raise ValueError(
@@ -316,7 +316,7 @@ class Multiscale(BaseAttrs):
                 raise ValueError(
                     "Transformations in coordinateTransformations must have an output."
                 )
-            
+
             if not hasattr(transformation.input, "name"):
                 raise ValueError(
                     "Input for coordinate transformations must provide"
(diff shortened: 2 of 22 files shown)


jo-mueller:
@dstansby I rebased this on #424 . Tests are green on my end and in the CI except for the one which loads remote data, which needs to be updated to a new 0.6.dev4 dataset.

Left to do:

The `Multiscales` class still has remnants of the coordinate system ordering in its code, particularly in the `_from_v05` class method. Once this is merged, I would:

- remove these remants -  could be as simple as changing some docstrings and refactor variables (i.e., `top_level_system` -> `output_coordinate_system`)
- move the `_from_v05` function to the a `_to_v06(...)` function in the v05 `Multiscales`, so that we're compatible with the overall `to_version` scheme.

dstansby:
Can you put a `pytest.mark.xfail` on the failing test, and a comment that it needs updating? Since it's only one test I'd rather do that and have the CI green before merging to main.

dstansby:
Also there are lots of pre-commit fixes that need making here.

jo-mueller:
pre-commit.ci autofix

jo-mueller:
pre-commit.ci autofix

jo-mueller:
@dstansby ok, looking better. The codecov fail is probably due to the added xfail in one of the tests.

src/ome_zarr_models/_v06/coordinate_transforms.py
```diff
@@ -70,16 +70,13 @@ class CoordinateSystemIdentifier(BaseAttrs):
     relative to the current Zarr group.
     """
 
-    name: str = Field(..., description="Coordinate system name.")
+    name: str | None = Field(default=None, description="Coordinate system name.")
     path: str | None = Field(
         default=None,
-        description="Path to Zarr group where the coordinate system is defined.",
+        description="Path to Zarr group.",
```
dstansby, line 76: Am I right in thinking that either `name` or `path` or both have to be set, ie they can't both be `null`? If so, there should be some validation that both are not `None` here as a pydantic class validator.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -64,9 +65,13 @@ def intrinsic_coordinate_system(self) -> CoordinateSystem:
...
         Notes
         -----
-        This is the last entry in the `coordinateSystems` attribute.
+        This is the `coordinateSystems` instance which is used output
+        of all multiscale transformations defined in `datasets`.
         """
-        return self.coordinateSystems[-1]
+        output = self.datasets[0].coordinateTransformations[0].output
+        if output is None:
+            raise ValueError("Output coordinate system is required")
```
dstansby:
```suggestion
            raise ValueError("Output coordinate system is not set in the metadata of the first dataset.")
```
To be a bit more specific about what the error is, and how the user could fix it.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -64,9 +65,13 @@ def intrinsic_coordinate_system(self) -> CoordinateSystem:
...
+        return next(cs for cs in self.coordinateSystems if cs.name == output.name)
```
dstansby, line 77:
```suggestion
        # First coordinate system that has same name as the output  system of the first dataset
        return next(cs for cs in self.coordinateSystems if cs.name == output.name)
```

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -148,9 +159,12 @@ def _ensure_same_output_cs_for_all_datasets(data: Self) -> Self:
         Also ensures that the dimensionality of each dataset match the one of the
         output coordinate system.
         """
-        output_cs_names = {
-            dataset.coordinateTransformations[0].output for dataset in data.datasets
-        }
+        output_cs = [
+            ds.coordinateTransformations[0].output
+            for ds in data.datasets
+            if ds.coordinateTransformations[0].output is not None
```
dstansby, line 168: It's a bit of a pain that this check has to be included, when the spec says that all of these transforms have to have their output set, but I can't think of a neat way to encode this in the types, so I think this is fine for now.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
         """
-        # TODO: this test is only for coordinate transformations that are not part of a
-        #  dataset. A second test for datasets should be added.
         if self.coordinateTransformations is None:
             return self
         cs_names = {cs.name for cs in self.coordinateSystems}
 
         # check input
         for transformation in self.coordinateTransformations:
+            if not hasattr(transformation.input, "name"):
```
dstansby: `transformation.input` will always have a `.name` attribute, no? Should the check here actually be `if transformation.input is None`?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
+                raise ValueError(
+                    "Input for coordinate transformations must provide"
+                    " a coordinate system name"
+                )
+
+            if not hasattr(transformation.output, "name"):
```
dstansby: Same comment as above, should this check instead be `if transformation.ouput is None`?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
+                raise ValueError(
+                    "Output for coordinate transformations must provide"
+                    " a coordinate system name"
+                )
             # TODO: add support for the input coordinate system being equal to the path
```
dstansby: Is this still a TODO, or is it fixed by this PR, or is it redundant with the current iteration of RFC5?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
             #  of the array data. See more:
             # https://imagesc.zulipchat.com/#narrow/channel/469152-ome-zarr-models-py/topic/validating.20paths
-            if transformation.input not in cs_names:
+            input_cs = transformation.input
+            if input_cs is not None and input_cs.name not in cs_names:
                 raise ValueError(
                     "Invalid input in coordinate transformation "
-                    f"'{transformation.name}': "
-                    f"{transformation.input}. Must be one of {cs_names}."
+                    f"'{input_cs}': "
```
dstansby:
```suggestion
                    f"'{transformation.name}': "
```
Shouldn't this still give the transformation name?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
+            if (
+                output_cs is not None
+                and not output_cs.path
+                and output_cs.name not in cs_names
+            ):
                 raise ValueError(
-                    "Invalid output in coordinate transformation: "
-                    f"{transformation.output}. Must be one of {cs_names}."
+                    "Invalid output in coordinate transformation "
+                    f"'{output_cs}': "
```
dstansby:
```suggestion
                    f"'{transformation.name}': "
```

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
+                    f"{output_cs.name}. Must be one of {cs_names}."
```
dstansby:
```suggestion
                    f"{output_cs.name}. Must be one of {cs_names} or the path field must be set."
```
Or some similar wording to indicate that another way to fix this issue is to set the `path`.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
                 )
+
+            # TODO: if output has path attribute, also check that
+            # the path is a multiscales group that hasa coordinate system
+            # with the specified name in transformation.output.name.
+            if output_cs is not None and output_cs.path:
```
dstansby:
```suggestion
            if output_cs is not None and output_cs.path is not None:
```
A bit safer to check `is not None`.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +288,69 @@ def check_dataset_transform_output(self) -> Self:
...
+                if output_cs.path.startswith(".."):
```
dstansby:
```suggestion
                if output_cs.path.contains(".."):
```
I don't know if this is what RFC5 says, but I think checking for a `..` anywhere in the path is a good idea instead of just at the beginning?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -378,11 +427,34 @@ def build(
 
     @model_validator(mode="after")
     def check_cs_input(self) -> Self:
+        """
+        Check the input of all multiscale transformations
+        """
         for transformation in self.coordinateTransformations:
-            if transformation.input != self.path:
+            # check that path field of input exists
+            if not hasattr(transformation.input, "path"):
```
dstansby: As above, this attribute should always exist. Should this check be `if transformation.input is not None`?

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -378,11 +427,34 @@ def build(
...
+        return self
+
+    @model_validator(mode="after")
+    def check_cs_output(self) -> Self:
+        """
+        Check the output of all multiscale transformations
+        """
+        for transformation in self.coordinateTransformations:
+            # check that output is a name (and not a path)
+            if not hasattr(transformation.output, "name"):
```
dstansby: As above, this attribute should always exist. Should this check be if `transformation.output is not None`?

tests/_v06/test_multiscales.py
```diff
@@ -427,10 +446,14 @@ def test_unique_system_names() -> None:
...
+                            output=CoordinateSystemIdentifier(name="out"),
                         ),
                     ),
                 ),
             ),
         )
+
+
+if __name__ == "__main__":
+    test_from_v05()
```
dstansby:
```suggestion
```

dstansby (changes requested):
I've had a look at the code and left some comments - I think there's a few bits that need changing, but nothing too major. Thanks a lot for making all these changes, almost there!

The graph code probably needs properly testing again (did you check if the graphs it produces look good?), but that can definitely happen after this PR.

dstansby:
(p.s., don't worry about codecov failing)

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: Yeah...you would probably need different types of transforms (i.e. a `MultiscalesTransform` of sorts), for which you could narrow down the typing a bit more, if you wanted to bake it into the types rather than the actual code.

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: In a multiscales, the `input` only needs a path (i.e., the path to a resolution level), but not a `name`.

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: right!

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: right!

jo-mueller: Oh wait. Actually both is needed. The input/output mustn't be `None`, and 

- `input.path` mustn't be `None`
- `output.name` mustn't be `None`

jo-mueller: 0c9e4f0fb468c2ae6d6c9cbb13d5fe76070ace67

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: See 0c9e4f0fb468c2ae6d6c9cbb13d5fe76070ace67

src/ome_zarr_models/_v06/coordinate_transforms.py
jo-mueller: 07ca957fcbd3254e832dd5d86e01ee910d8a51f2

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: There is a model validator that makes sure that all multiscale transforms output to the same coordinate system (`_ensure_same_output_cs_for_all_datasets`). So...although this property only pulls the coordinate system of the first multiscales transform, if this fails then the error is likely not just in the first scale level

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: Same as https://github.com/ome-zarr-models/ome-zarr-models-py/pull/428#discussion_r3161488383

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: See fd3b7cc60c2d585328135d87ca90f7de1f4fdbc0

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: See fd3b7cc60c2d585328135d87ca90f7de1f4fdbc0

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: See fd3b7cc60c2d585328135d87ca90f7de1f4fdbc0

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: See fd3b7cc60c2d585328135d87ca90f7de1f4fdbc0

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: Jup. removed in fdc824a4b5885e2c2a0fb35e8b3c3a42f09f109c

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: Oh wait. That's not in the multiscales datasets. Both `input` and `output` must exist, and both must have the `name` set, which will reference a coordinate system either in the same metadata document or elsewhere.

jo-mueller:
pre-commit.ci autofix

jo-mueller:
@dstansby that should do it :)

jo-mueller:
@d-v-b @lorenzocerrone @imagejan with @dstansby do you know who will maintain this repository? Minimally who could or would have a look or merge this PR?

dstansby:
👋 I'm just back from holiday. I intend to carry on maintaining this repository, at least in the short term, though others on [the developers team](https://github.com/orgs/ome-zarr-models/teams/ome-zarr-models-developers) are of course welcome to review and mege PRs.

I'll try and take a look at this PR later this week.

src/ome_zarr_models/_v06/multiscales.py
```diff
@@ -274,36 +291,77 @@ def check_dataset_transform_output(self) -> Self:
...
+            if transformation.input is None:
+                raise ValueError(
+                    "Transformations in coordinateTransformations must have an input."
+                )
+            if transformation.output is None:
+                raise ValueError(
+                    "Transformations in coordinateTransformations must have an output."
+                )
+
+            if not hasattr(transformation.input, "name"):
```
dstansby: My original comment still stands; this will always evaluate to `False`, because `CoordinateSystemIdentifier` always has a `name` attribute, even if it takes a default value:

```python
from ome_zarr_models._v06.coordinate_transforms import CoordinateSystemIdentifier

csi = CoordinateSystemIdentifier(path="/path")
print(csi)  # name=None path='/path'
print(hasattr(csi, "name"))  # True
```

So I think these `if` statements need changing to e.g., `if transformation.input.name is None` instead.

dstansby (changes requested):
All good, apart from one point in my original review that I still think needs addressing.

src/ome_zarr_models/_v06/multiscales.py
jo-mueller: @dstansby fixed in c88f4a0565c07079514185d45e66d076c0ab2866

dstansby (approved):

PR #432, opened 2026-05-12
Refactor: Add 06 version converter

jo-mueller:
Built on #428 . This PR refactors the already existing conversion classmethod `_from_v05` into the existing conversion framework orchestrated by the `to_version` functions that were introduced in #398

PR #433, opened 2026-05-13
refactor: move 0.5->0.6 conversion logic into 05 Multiscales class

jo-mueller:
Clean version of #432

Built on #428 and #432 . This PR refactors the already existing conversion classmethod _from_v05 into the existing conversion framework orchestrated by the to_version functions that were introduced in https://github.com/ome-zarr-models/ome-zarr-models-py/pull/398


jo-mueller:
pre-commit.ci autofix

jo-mueller:
@dstansby I think this should be a simple one. I'd very much like to have a `_to_v05` downgrade method in the 0.6 Multiscale metadata object, but that'll be a follow-up PR.

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -56,18 +63,92 @@ class Multiscale(BaseAttrs):
...
+        from ome_zarr_models._v06.multiscales import (
+            Dataset as DatasetV06,
+        )
+        from ome_zarr_models._v06.multiscales import (
+            Multiscale as MultiscaleV06,
+        )
+
+        ms_v06 = MultiscaleV06(
+            datasets=tuple(
+                DatasetV06(
```
dstansby, line 129: To separate the logic a bit (and mirror `Multiscale._to_v06()`, could there instead be a `Dataset._to_v06()` method added that's called here?

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -56,18 +63,92 @@ class Multiscale(BaseAttrs):
     name: JsonValue | None = None
     type: JsonValue = None
 
-    def to_version(self, version: Literal["0.4"]) -> MultiscaleV04:
+    def to_version(
+        self, version: Literal["0.4", "0.6"]
```
dstansby: There should be a way here of specifying the extra information that gets added to a 0.6 multiscales. In the original code this was:

- The name of the intrinsic system (currently hardcoded to "physical")
- The 'top level' coordinate system

There might be others too? These can take sensible defaults, but I think it's important to give the user the option of specifying them.

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -56,18 +63,92 @@ class Multiscale(BaseAttrs):
...
             raise ValueError(f"Unsupported version conversion: 0.5 -> {version}")
 
+    def _to_v06(self) -> MultiscaleV06:
+        from ome_zarr_models._v06.coordinate_transforms import (
+            Axis,
+            CoordinateSystem,
+            CoordinateSystemIdentifier,
+        )
+        from ome_zarr_models._v06.multiscales import (
+            Dataset as DatasetV06,
```
dstansby:
```suggestion
            Dataset as DatasetV06
```
If you remove the trailing commas, `ruff` should format this on one line instead of three.

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -56,18 +63,92 @@ class Multiscale(BaseAttrs):
...
+        )
+        from ome_zarr_models._v06.multiscales import (
+            Multiscale as MultiscaleV06,
```
dstansby:
```suggestion
            Multiscale as MultiscaleV06
```

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -338,3 +419,36 @@ def _ensure_transform_dimensionality(
             )
             raise ValueError(msg)
         return transforms
+
+
+def _v05_transform_to_v06(transform: ValidTransform) -> ScaleV06 | SequenceV06:
+    from ome_zarr_models._v06.coordinate_transforms import (
+        Scale as ScaleV06,
```
dstansby:
```suggestion
        Scale as ScaleV06
```

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -338,3 +419,36 @@ def _ensure_transform_dimensionality(
...
+    )
+    from ome_zarr_models._v06.coordinate_transforms import (
+        Sequence as SequenceV06,
+    )
+    from ome_zarr_models._v06.coordinate_transforms import (
+        Translation as TranslationV06,
```
dstansby:
```suggestion
        Translation as TranslationV06
```

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -338,3 +419,36 @@ def _ensure_transform_dimensionality(
...
             raise ValueError(msg)
         return transforms
+
+
+def _v05_transform_to_v06(transform: ValidTransform) -> ScaleV06 | SequenceV06:
+    from ome_zarr_models._v06.coordinate_transforms import (
+        Scale as ScaleV06,
+    )
+    from ome_zarr_models._v06.coordinate_transforms import (
+        Sequence as SequenceV06,
```
dstansby:
```suggestion
        Sequence as SequenceV06
```

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -338,3 +419,36 @@ def _ensure_transform_dimensionality(
...
+    from ome_zarr_models.common.coordinate_transformations import (
+        VectorScale,
+        VectorTranslation,
+    )
+
+    # Scale (always present)
+    if isinstance(transform[0], VectorScale):
+        scale = ScaleV06(scale=tuple(transform[0].scale))
+    else:
+        scale = ScaleV06(path=transform[0].path)
```
dstansby, line 474: 0.6 doesn't support paths any more right? If so, this needs to be changed to an error (and same below with the translation).

jo-mueller: Jep. I thought about including that in this PR, but it comes with quite a few changes that aren't technically related to this PR, so I'm saving it for later.

src/ome_zarr_models/v05/multiscales.py
jo-mueller: I thought about it, but that would require the whole `to_version` machinery for all `Dataset` classes. And I don't think a dataset instance is ever used without a `Multiscales` instance, so I think it's fair to do it only on the level of `multiscales`?

src/ome_zarr_models/v05/multiscales.py
jo-mueller: Fixed in dcd2363176e5c936d045d144b5c0af1535e7c8a3. I left it in because in case somebody wanted to convert their data, they probably didn't care about the coordinate system's name in the 0.5 metadata, they wouldn't care much about it in the 0.6 metadata either.

Now it's possible to set on conversion 👍

jo-mueller:
pre-commit.ci autofix

dstansby:
I've opened a PR to this branch to improve the typing: https://github.com/jo-mueller/ome-zarr-models-py/pull/1 - because there's one function for different versions of OME-Zarr conversion, it's a bit annoying to type, but I think I came up with an okay solution. Improvements very welcome though...!

The other thing that this PR should update is the user example at `docs/convert.py`. It shouldn't be too hard to update since the new API is simple, so it would be nice to update at the same time as this change to make sure that this all works nicely from a user perspective.

jo-mueller:
By improving the typing you mean being able to call `to_version(...)` without having to worry about typing because the return type may or may not be one out of `MultiscaleV04 | MultiscaleV06` etc?

dstansby:
I added typing overloads to `to_version(...)` that help by 1) narrowing the return type based on the value of the `version` argument 2) narrowing the allowed signatures based on the value of the `version` argument.

jo-mueller:
Cool, didn't know it was possible to narrow return types based on input arguments 😲

dstansby:
Ugh, that breaks the doc build, sorry. Things left to do are:

- Fix the doc build, I think by adding back explicit arguments instead of generic `kwargs` to the actual method signature.
- Update the doc example in `docs/convert.py`

jo-mueller:
@dstansby ok I think it's good to go :)

dstansby:
Please could you also update the doc example in docs/convert.py, so there's a user facing guide to using the converter. It doesn't have to be a long example, but I think it would be nice to show the functionality and explain at the same time what the new keyword arguments mean for the 0.6 metadata when 'upgrading' from 0.5.

jo-mueller:
@dstansby added in e72cea3b2a75258ade8c5e7395761b5ef470e90e

jo-mueller:
pre-commit.ci autofix

dstansby:
🚢

PR #434, opened 2026-05-13
Add 0.6 -> 0.5 down converter

jo-mueller:
@dstansby follow-up to #432 and the last puzzle piece that would make ozmp complete, at least for my purposes ;)

The PR adds a converter function `_to_v05` that converts everything to 0.5 with the following conditions:
- `coordinateTransformations` are omitted - there's a subset of Scale and Translation that *could* be ported, but that seems like a lot of additional complexity at this point.
- The converter doesn't check whether multiple coordinate systems exist in the scope of the `Multiscales` object - it just picks the intrinsic one and takes the `axes` from there.


jo-mueller:
pre-commit.ci autofix

jo-mueller:
pre-commit.ci autofix

jo-mueller:
@dstansby I think this is ready for a first look, too. Once #438 is merged, I'll rebase or port over to a new PR to clean up the commit history.

PR #438, opened 2026-05-19
Simplify scale translate transforms

### LINKED ISSUE #435: Simplify 0.6 Scale and Translation classes?
In the 0.6 transforms, `Scale` and `Translation` still have the [path fields](https://github.com/ome-zarr-models/ome-zarr-models-py/blob/791659604edef2a3b552819ddb30dd945886452f/src/ome_zarr_models/_v06/coordinate_transforms.py#L311), but those aren't in the latest version of the spec anymore. 

This allows these classes to be a fair bit shorter, because a lot of the logic in there just handles the case of whether `path` or `scale` is present. I.e., the `scale_vector` and `translation_vector` properties are no longer needed, imho?

jo-mueller:
Fixes #435 

Based on #433 

This PR removes the `path` fields from the scale and translation transforms.

Re-opened from #437 for clean commit history.

I slipped in another thing I noticed: In the `SceneAttrs` class, `coordinateSystems` cannot be set to `None`, but the spec allows scene metadata without coordinate systems.

### DIFF
--- a/src/ome_zarr_models/_v06/scene.py
+++ b/src/ome_zarr_models/_v06/scene.py
@@ -18,7 +18,7 @@ from ome_zarr_models._v06.image import Image
 
 class SceneAttrs(BaseModel):
     coordinateTransformations: tuple[AnyTransform, ...] = Field(default=())
-    coordinateSystems: tuple[CoordinateSystem, ...] = Field(default=())
+    coordinateSystems: tuple[CoordinateSystem, ...] | None = Field(default=None)
 
 
 class BaseSceneAttrs(BaseOMEAttrs):
@@ -132,8 +132,9 @@ class Scene(BaseGroupv06[BaseSceneAttrs]):
         graph = TransformGraph()
 
         # Coordinate systems
-        for system in self.ome_attributes.scene.coordinateSystems:
-            graph.add_system(system)
+        if self.ome_attributes.scene.coordinateSystems is not None:
+            for system in self.ome_attributes.scene.coordinateSystems:
+                graph.add_system(system)
         # Coordinate transforms
         for transform in self.ome_attributes.scene.coordinateTransformations:
             graph.add_transform(transform)
--- a/src/ome_zarr_models/v05/multiscales.py
+++ b/src/ome_zarr_models/v05/multiscales.py
@@ -470,8 +470,6 @@ def _v05_transform_to_v06(transform: ValidTransform) -> ScaleV06 | SequenceV06:
     # Scale (always present)
     if isinstance(transform[0], VectorScale):
         scale = ScaleV06(scale=tuple(transform[0].scale))
-    else:
-        scale = ScaleV06(path=transform[0].path)
 
     if len(transform) == 1:
         return scale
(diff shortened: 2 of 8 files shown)


src/ome_zarr_models/_v06/scene.py
```diff
@@ -18,7 +18,7 @@
 
 class SceneAttrs(BaseModel):
     coordinateTransformations: tuple[AnyTransform, ...] = Field(default=())
-    coordinateSystems: tuple[CoordinateSystem, ...] = Field(default=())
+    coordinateSystems: tuple[CoordinateSystem, ...] | None = Field(default=None)
```
dstansby, line 21: Why does this have a default now?

dstansby: Sorry, saw your comment above! All good

dstansby:
pre-commit.ci autofix

dstansby:
Looks like there's still some typing errors

jo-mueller:
@dstansby fixed 👍

src/ome_zarr_models/v05/multiscales.py
```diff
@@ -470,8 +470,6 @@ def _v05_transform_to_v06(transform: ValidTransform) -> ScaleV06 | SequenceV06:
     # Scale (always present)
     if isinstance(transform[0], VectorScale):
         scale = ScaleV06(scale=tuple(transform[0].scale))
-    else:
-        scale = ScaleV06(path=transform[0].path)
```
dstansby, line 474: If a transform with a path is passed to this function, shouldn't it error because it's not possible to convert to 0.6?

jo-mueller: strange that mypy hasn't picked that up 🤔

jo-mueller: I think it's because the converter just does conversion from `VectorScale` to `ScaleV06`. In the 0.5 model, `VectorScale` and `PathScale` are two different concepts. I guess to make it more bulletproof one would have to add an additional check on whether the input scale transform is of type `VectorScale` or `PathScale`. In the latter case, we'd have to load the scale values from zarr and add them to the `ScaleV06` class.

jo-mueller: Would you mind if I created an issue for that and fixed it in a follow-up PR?

dstansby: I don't think we do any loading of values from `PathScale` anyway, so definitely 👍 to not load the values and just error if given a `PathScale` (or `PathTranslation`)

dstansby: If you can fix it in a follow up PR even better 😄 but I don't think it's neccesary for 0.6 support.

jo-mueller: Right? So far I've yet to see ome-zarrs that actually make use of storing scale transforms in zarr arrays 🙊

dstansby: This still needs to error in this PR

jo-mueller:
pre-commit.ci autofix

dstansby (approved):

PR #440, opened 2026-05-20
A cleaner way to specify extra 0.6 metadata when converting multiscales

dstansby:
This grew out of me wanting to understand what extra metdata can be added to 0.6 multsicales, that can't be captured in a 0.5 multiscales.

It adds a new class, `Extrav06Metadata`, that encapsulates all the extra bits of metadata you can add.

I haven't quite finished implementing all the bits (discrete and longNames), but I wanted to open this early to see if @jo-mueller (or anyone else!) has any comments/thoughts on the design and approach here.|

### DIFF
--- a/docs/convert.py
+++ b/docs/convert.py
@@ -1,55 +1,35 @@
 # This page explains how to convert metadata from one OME-Zarr version to another.
-#
-# ## 0.5 to 0.6
-#
-# The main addition to OME-Zarr in version 0.6 is coordinate systems.
-
 
 import zarr.storage
 from rich import print
 
-import ome_zarr_models._v06
 import ome_zarr_models.v05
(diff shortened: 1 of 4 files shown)


jo-mueller:
Hm...my feeling is that I would put it more on the shoulders of a good API (i.e., consumers or users of ome-zarr-models-py) that orchestrates 0.5->0.6 conversion to come up with a good way to set these terms, rather than adding additional code here that will have to wrapped by a consumer implementation.

Right now I don't see much extra value in creating `Extrav06Metadata` instances over converting to 0.6 metadata and updating the respective fields (i.e., `discrete`, `longName`, etc) post hoc using `model_copy(update=...)`.

dstansby:
That makes sense - I'll close this then, and open a new PR that adds documentation saying users should manually update the new fields if they want to add the new information.

PR #442, opened 2026-05-22
Prepare docs for experimental 0.6 support

dstansby:
This is some housekeeping to put files in the right place for version 0.6.

- Rename the 0.6 submodule from `_v06` > `v06`. This way users won't have to update imports between now and the final support for 0.6.
- Rename the doc files from `dev` to `v06`
- Add warnings in the API docs


dstansby:
I'm going to merge this, because otherwise the file renaming will be a headache of merge conflicts if people are working off main.

PR #447, opened 2026-05-22
Only allow 0.6dev4 as version number for 0.6 

dstansby:
Since there have been major changes throughout 0.6 development, pin the allowed version numbers to an exact development version. @jo-mueller what do you think? This prevents loading older or newer data, but I think that's good because we don't want user (or developer) confusion around exactly what version of the specification ome-zarr-models represents.

Depends on https://github.com/jo-mueller/ngff-rfc5-coordinate-transformation-examples/pull/45

### DIFF
--- a/src/ome_zarr_models/v06/base.py
+++ b/src/ome_zarr_models/v06/base.py
@@ -17,9 +17,8 @@ class BaseOMEAttrs(BaseAttrsv3):
     Base class for OME-Zarr 0.6 attributes.
     """
 
-    # TODO: added 0.6.dev* to allow for RFC5 testing.
-    # Remove this before final release!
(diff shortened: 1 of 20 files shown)


jo-mueller:
I think it would be nice for the conversion to also write that number? i *think* (untested) if you do `multiscale.to_version("0.6") after this PR, what you get is an error because 0.6 is not allowed?

dstansby:
There's now a test at https://github.com/ome-zarr-models/ome-zarr-models-py/pull/450/changes#diff-1209164a3b942213535dc113f044dd8cf0086c8d133f34cb05a8f4032e4417f3R277, which shows that "0.6.dev4" is written. Other than that, do you think the changes here are a good idea?

jo-mueller (approved):

PR #448, opened 2026-05-22
Add transformation conformance tests

clbarnes:
Uses conformance tests defined here on CI https://github.com/clbarnes/ome_zarr_transformations_conformance#

Most tests fail at the moment as a few things are yet to be implemented

- transform graphs finding inverse paths
- path-form affines/ rotations
- byDimension transformations

Because of this, I have `continue-on-error: true` on that step, which GHA lists as a success, which makes it not very helpful 🙃 

Note that the conformance tests themselves are not yet complete too.

This exercise has allowed me to fix a bunch of issues with the conformance tests, though!


dstansby:
Very nice! Could you open separate issues for the stuff that isn't working yet?

clbarnes:
NB constraining the version specifier to exactly 0.6.dev4 breaks some tests; could it accept (but maybe warn) on "0.6"?

dstansby:
> NB constraining the version specifier to exactly 0.6.dev4 breaks some tests; could it accept (but maybe warn) on "0.6"?

👍 - it might be nice to 'fix' the version number being passed in by the conformance tests as 0.6.dev4 though too? I thought the idea was to give pre-release OME-Zarr data 0.x.devz version numbers to distinguish it from data written the final released specification.

clbarnes:
> it might be nice to 'fix' the version number being passed in by the conformance tests as 0.6.dev4 though too?

On the one hand, yes; on the other, I sort of can't be bothered digging through multiple repos to find-and-replace every time a new dev version comes out.

jo-mueller:
@clbarnes just looking at this more carefully: SInce the transform graph functionality is not living here anymore, but will come to ome-zarr-py in https://github.com/ome/ome-zarr-py/pull/612 - should this conformance test live at ome-zarr-py then?

See also https://github.com/ome/ome-zarr-py/issues/592

clbarnes:
Yes, I think this should go in ome-zarr-py.

PR #458, opened 2026-05-27
Allow "0.6" as a version number

### LINKED ISSUE #457: Allow version number "0.6" in dev 0.6 implementation
I changed the 0.6 versioning to only accept "0.6.dev4" as the version number, but it should be easy enough to allow "0.6" as input (see https://github.com/ome-zarr-models/ome-zarr-models-py/pull/448#issuecomment-4554569695 by @clbarnes for the request).

The implementation should probably have a pydantic before validator on version that accepts "0.6", but converts it to "0.6.dev4".

dstansby:
Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/457

### DIFF
--- a/src/ome_zarr_models/v06/base.py
+++ b/src/ome_zarr_models/v06/base.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING, Generic, Literal, Self, TypeVar, Union
 
 import pydantic_zarr
 import pydantic_zarr.v3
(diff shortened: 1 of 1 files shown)


clbarnes:
Looks good! One could parse the version string completely with something like [parver](https://pypi.org/project/parver/) as we're de facto using PEP440 version strings, but I don't anticipate much need for implementations to support intermediate dev versions once 0.6 lands.

clbarnes (approved):

PR #462, opened 2026-06-01
Add method to get transformation between two systems.

### LINKED ISSUE #461: Support getting a full transformation between two nodes in a graph
Currently it's only possible to get the path between two nodes in a transformation graph (ie., the list of nodes in the path). There should be a high level API on the graph that returns the full transformation instead.

dstansby:
Fixes https://github.com/ome-zarr-models/ome-zarr-models-py/issues/461

cc @clbarnes, I don't know if this is useful for conformance testing?

### DIFF
--- a/src/ome_zarr_models/_utils.py
+++ b/src/ome_zarr_models/_utils.py
@@ -5,7 +5,9 @@ Private utilities.
 from __future__ import annotations
 
 import heapq
+import itertools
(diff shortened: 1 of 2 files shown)


clbarnes:
For the equivalent method in transformnd, I also cached the result with `@lru_cache()` (clearing the cache whenever the graph was mutated).

PR #464, opened 2026-06-17
feat: add converter from 0.4 -> 06

jo-mueller:
I realized that it's possible to convert from 0.4 -> 0.5 and also from 0.5 -> 0.6, but it's **not** possible to do 0.4 ->0.6, which is a very straightforward (and useful thing) to do by just wrapping 0.4 -> 0.5 -> 0.6 into a single conversion call.

This PR adds that feature. Not sure we need tests there because both the 0.4->05 and 0.5 -> 0.6 conversions are already covered.


jo-mueller:
pre-commit.ci autofix

Issue #463, opened 2026-06-01
Replace custom coordinate graph ipmlementation with `transformnd.TransformGraph`?


dstansby:
I didn't realise until reading https://github.com/ome-zarr-models/ome-zarr-models-py/pull/462#issuecomment-4590987955 that `transformnd` [implements coordinate graphs too](https://transformnd.readthedocs.io/en/stable/transformnd.html#TransformGraph.get_sequence). Perhaps instead of rolling our own transform graph logic, we should (optionall) depend on `transformnd` for this.

See also https://github.com/ome-zarr-models/ome-zarr-models-py/issues/363 for related discussion about optionally depending on `transformnd` for actual transform logic.

If we do this there would be a nice separation of responsibilities - `ome-zarr-models` is responsible for parsing and validating OME-Zarr metadata, and then hands off any logic for actually transforming and working with the transformations to `transformnd`. It would slim down `ome-zarr-models`, and I think we could also contribute good suggestions for new features to `transformnd` too if we used it. This was suggested by @jo-mueller too: https://github.com/ome-zarr-models/ome-zarr-models-py/issues/363#issuecomment-4440892404

The downside would be potential user confusion because of an optional dependency, and heavier dependencies (slightly mitigated if it's optional).

@jo-mueller @clbarnes @LucaMarconato do you have any thoughts on this?

clbarnes:
I'd be happy with this - was going to suggest it myself as transformnd also provides more performance-oriented transformations. A couple of other thoughts:

- transformnd's deps are basically numpy and networkx which are almost certainly in every scientific venv anyway
- transformnd currently doesn't support
  - transforms which change dimensionality (affines and coordinates transforms can do this);o n the roadmap and I don't think would be too hard to change
  - vector field transforms (coordinates/ displacements); shouldn't be too bad either
- [ome-zarr-transformnd](https://github.com/clbarnes/ome-zarr-transformnd) handles going from OME-Zarr metadata (as dicts and zarr-python types) to a transformnd transform graph; I'd be happy to build it on top of ome-zarr-models instead
- does ome-zarr-models-py actually need to depend on transformnd, or should they instead both be depended on by a library like ome-zarr-transformnd and/or ome-zarr-py which people would use for actually doing the transformations in context? Is constructing and interrogating a graph, and using it to apply transformations, actually required for the validation ome-zarr-models-py does?

jo-mueller:
[...]

PR #468, opened 2026-06-25
Add to_transformnd method to all transforms

dstansby:
A pre-requisite for https://github.com/ome-zarr-models/ome-zarr-models-py/issues/463. @clbarnes, is there any chance you could have a quick look at this to check that it seems right?

### DIFF
--- a/src/ome_zarr_models/v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/v06/coordinate_transforms.py
@@ -12,7 +12,7 @@ from ome_zarr_models.common.validation import unique_items_validator
 
 if TYPE_CHECKING:
     import transformnd
-    import transformnd.transforms  # noqa: TC004
+    import transformnd.transforms
 
     from ome_zarr_models.v05.axes import Axis as Axisv05
 
@@ -587,6 +587,8 @@ class Sequence(Transform):
     transformations: tuple[AnyTransform, ...]
 
     def to_transformnd(self) -> transformnd.TransformSequence:
+        import transformnd
+
         return transformnd.TransformSequence(
             transforms=[t.to_transformnd() for t in self.transformations]
         )
--- a/src/ome_zarr_models/v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/v06/coordinate_transforms.py
@@ -415,7 +415,10 @@ class Affine(Transform):
     def to_transformnd(self) -> transformnd.transforms.Affine:
         import transformnd.transforms
 
-        return transformnd.transforms.Affine(matrix=self.affine_matrix)
+        bottom_row = [0] * (self.ndim + 1)
+        bottom_row[-1] = 1
+        matrix = np.vstack([self.affine_matrix, bottom_row])
+        return transformnd.transforms.Affine(matrix=matrix)
 
     @classmethod
     def _from_matrix_vector(
--- a/src/ome_zarr_models/v06/coordinate_transforms.py
+++ b/src/ome_zarr_models/v06/coordinate_transforms.py
@@ -11,6 +11,9 @@ from ome_zarr_models.base import BaseAttrs
 from ome_zarr_models.common.validation import unique_items_validator
 
 if TYPE_CHECKING:
+    import transformnd
+    import transformnd.transforms  # noqa: TC004
+
     from ome_zarr_models.v05.axes import Axis as Axisv05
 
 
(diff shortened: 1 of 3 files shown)


src/ome_zarr_models/v06/coordinate_transforms.py
```diff
@@ -213,6 +222,11 @@ class Identity(Transform):
 
     type: Literal["identity"] = "identity"
 
+    def to_transformnd(self) -> transformnd.transforms.Identity:
+        import transformnd.transforms
+
+        return transformnd.transforms.Identity()
```
clbarnes, line 228: This shouldn't work - transformnd requires a dimensionality for all transforms (infers them most of the time but there's an explicit `ndim` arg for Identity). Not sure why the tests aren't picking that up.

src/ome_zarr_models/v06/coordinate_transforms.py
```diff
@@ -383,6 +412,11 @@ class Affine(Transform):
         default=None, description="Path to affine matrix stored as a Zarr array."
     )
 
+    def to_transformnd(self) -> transformnd.transforms.Affine:
+        import transformnd.transforms
+
+        return transformnd.transforms.Affine(matrix=self.affine_matrix)
```
clbarnes: I think this won't work, because OME-Zarr doesn't store the bottom row of the affine, but transformnd requires it.

src/ome_zarr_models/v06/coordinate_transforms.py
clbarnes: When I've been fiddling with a bridge between OME-Zarr metadata and transformnd it has been a pain to deal with Identity for this reason, you end up needing the context of coordinate systems (possibly from multiple metadata documents) and the containing transform if you're inside a bydimension or sequence.

Transformnd originally didn't require explicit in/out dimensionality but it made some other things much more complicated.

dstansby: Woops, I fixed the tests not running. This is going to be a bit of a pain, I'll see if there's a nice way to set ndim from other context, but I'm slightly worried that if there's a `Identity` without any other transforms present it's going to be hard to infer `ndim` because it will need to come from Zarr array metadata that is not in the same place as OME-Zarr metadata.

src/ome_zarr_models/v06/coordinate_transforms.py
dstansby: Thanks - should be fixed now.

LucaMarconato:
Shouldn't the conversion to `transformnd` transformations be responsibility of a library that imports both `ome-zarr-models-py` and `transformnd`, sufh as `ome-zarr-transforms-nd`?

clbarnes:
Yes, that's the other possibility. While it is convenient for OZM to offer this functionality, maybe it should be handled by a library which has knowledge of the full arrangement so you have the context available. I'm also aware that this is the exact opposite of the argument I'm making elsewhere in the projectAxis discussion 😅.

dstansby:
I hadn't thought of that option - @LucaMarconato is your suggestion that `ome-zarr-models` removes all functionality apart from metadata reading and validating, and leaves anything else (e.g., transform graphs, actually doing transformations) to other libraries? I think I would be fairly on board with that, but we should probably add "other operations" as out of scope formally in the documentation if we want to go down that route.

If you broadly agree with the above proposal, could you open an issue so others can comment, and we could take it from there?

LucaMarconato:
Almost, there would still 2 ways forward:
1. keep transformnd as a test dep, and to the testing of applying the transformations in ome-zarr-models-py using this dependency. The conversion to a transformnd transformation would not be done as an abstract method, but form a helper function living in conftest.py
2. remove all the testing of applying the transformations and using the external conformance testing machinery from @clbarnes 

The point 1. is quicker to implement/requires minimal changes, point 2. is a more structural refactoring but could pay off in terms of longer term maintenance. I don't have a strong preference for any of the 2.

LucaMarconato:
Done here! https://github.com/ome-zarr-models/ome-zarr-models-py/issues/469

jo-mueller:
> Shouldn't the conversion to `transformnd` transformations be responsibility of a library that imports both `ome-zarr-models-py` and `transformnd`, sufh as `ome-zarr-transforms-nd`? 

+1 from my side. In the implementations I use, everytging around metadata modeling and stuff that needs the array data is very strongly separated. I'm doing exactly what @LucaMarconato describes on my development head for scenes in ome-zarr-py

dstansby:
As per discussion above and https://github.com/ome-zarr-models/ome-zarr-models-py/issues/469, I'm going to close this.

Issue #469, opened 2026-06-25
Move ⁠`to_transformnd` conversion logic out of `⁠ome-zarr-models-py` into an upstream library


LucaMarconato:
Opened here as requested from @dstansby when discussing PR #468 (the PR proposes adding a `to_transformnd` method to all coordinate transformation classes in `ome-zarr-models-py`).

I suggest that managing conversions from OME-Zarr metadata to `transformnd` objects should not be handled by `ome-zarr-models-py` classes. This aligns with keeping the library scope focused on reading and validating OME-Zarr metadata, while delegating transformation operations to libraries that have complete contextual knowledge of how transformations are applied.

### Two possible approaches

1. **Minimal change**: Keep `transformnd` as a test-only dependency in `ome-zarr-models-py`, with `to_transformnd` as a helper function for the tests (`conftest.py`), not a method. This would allow for testing the correct application of transformations within this repository.
2. **Delegate to external libraries**: Remove `to_transformnd` methods entirely from `ome-zarr-models-py` and any functionality that requires knowledge of the transformations graph. Testing the correctness of the application of transformations would be conducted using the conformance testing machinery from @clbarnes (https://github.com/clbarnes/ome_zarr_transformations_conformance).

A drawback of both approaches is that one would not be able to validate the connectedness of the graph within this repository. This would need to be done in a downstream repository. Small "conflict of interest" disclosure: I would not mind removing the validation for the connectedness of the graph from this repository since I voiced concerns about the graph connectedness constraint (https://github.com/ome/ngff-spec/issues/155). However, this can be revisited in the context of NGFF 1.0 (or 0.7). We can adhere to the connectedness constraint and potentially relax it later.

dstansby:
I'm 👍 on removing transform logic, including `transformnd` adaptors, from ome-zarr-models. My main reasoning is that folks actively using ome-zarr-models are happy for this to happen, and it will keep ome-zarr-models nice and slim. I'll go ahead and open a PR that removes transformation logic, and folks can give feedback on that.

jo-mueller:
@LucaMarconato can this be closed? I'm not sure whether ozmp to transformnd conversion needs to live in its own project; I would kind of expect this functionality to be implemented separately in projects that make use of both transformnd and ozmp, especially since that requires some IO.

PR #470, opened 2026-07-01
Remove transform graph logic

dstansby:
See discussion in https://github.com/ome-zarr-models/ome-zarr-models-py/issues/469.

cc @jo-mueller and @LucaMarconato and @clbarnes just to double check that we're all on the same page about removing transform graph logic - my thinking is we could make a new little package at https://github.com/ome-zarr-models/ome-zarr-transform-graph for creating transform graphs with `transformnd` from ome-zarr data. Very open to other suggestions if others want to implement this functionality elsewhere though!


LucaMarconato:
Thanks @dstansby! Regarding the transformation graph, please notice that `ome-zarr-transformnd` (and maybe also `ome-zarr-py`?) contains code to build the `transformnd` graph from `ome-zarr-models-py` classes. 

Also in case you missed (see Zulip here [#ngff-coordinates-devs > new repository](https://imagesc.zulipchat.com/#narrow/channel/468962-ngff-coordinates-devs/topic/new.20repository/with/607102280)), we are currently expanding the experiments we did at the hackathon in https://github.com/BioImageTools/ngff-transformations into a new repo https://github.com/Spatial-Innovation-Team/transfo. 

The repo, which at this stage is only experimental code, includes an initial sketch of building the `transformnd` graph from `ome-zarr-models-py` classes and an [initial prototype of the xarray bridge](https://github.com/Spatial-Innovation-Team/transfo/blob/f2f17a53101abfd387f89a8ab806d9b3b1d95948/sandbox/ngff_transform_demo.py). It will also maybe contain the `spatialdata` transform classes (their focus is in-memory ergonomics). After we are done with our experiments we'll either contribute code around (some in `spatialdata`, some upstream), either we will polish the repo into a proper package (and change the name to a better one).

PR #465, opened 2026-06-17
feat: add projectAxis transform

jo-mueller:
PR mirroring https://github.com/ome/ngff-spec/pull/130

This adds the model class to represent the to-be-introduced `projectAxis` transform

PR #471, opened 2026-07-22
Project axis field requirements

jo-mueller:
Follow-up to #465 - noticed that the implementation would always required both `droppedInputs` and `createdOutputs` to be present, which is not the case. I made both fields optional and added a validator + tests to make sure that one of these fields is always present.


jo-mueller:
pre-commit.ci autofix

PR #476, opened 2026-08-24
feat: remove `version` key in plate/well metadata

jo-mueller:
To whom it may concern: This PR corrects some of the classes around Plates and Wells in the v06 metadata. I think already in v05, we didn't need the `version` key under the `plate` or `well` fields in the metadata, but it was still written (schemas and spec were a bit out of sync there).

Correct:

```json
"ome": {
  "version": ...,
  "plate": {...}
  }
```

Semi-incorrect:

```json
"ome": {
  "version": ...,
  "plate": {
    "version": ...
    }
  }
```

In v06, this is corrected, so we can simplify some things here.

cc @will-moore @dstansby @kevinyamauchi @jni

With this merged, I think I would cut a release for 1.9.0 instead of the pre-releases.

### DIFF
--- a/src/ome_zarr_models/v06/plate.py
+++ b/src/ome_zarr_models/v06/plate.py
@@ -2,8 +2,6 @@
 For reference, see the [plate section of the OME-Zarr specification](https://ngff.openmicroscopy.org/0.5/index.html#plate-md).
 """
 
-from pydantic import Field
-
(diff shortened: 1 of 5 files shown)


will-moore:
👍  for the cleanup - Code looks fine but I'm not familiar with it so can't review properly

